PR #11, opened 2025-06-27
Feature/mx 1893 igs access platform resource mandatory

erichesse:
### PR Context

- assets PR 245

### Added

- transform mandatory fields of IGS access platform and resources


esinsj (commented):
Looks good :)

esinsj (approved):

PR #25, opened 2025-07-08
mx_1938_avoid_endnote_identifier_collisions'

erichesse:
### PR Context
- address issue more comprehensively in MX-1938

### Fixed

- avoid identifier collision in endnote organizations and persons


esinsj (approved):
top

PR #26, opened 2025-07-08
feature/mx-1943-fix-endnote

cutoffthetop:
### Changes
- ignore records with 50 or more authors
- improve reading content from extracted raw data
- avoid repeated loading of the same organizations.

### DIFF
--- a/mex/extractors/endnote/extract.py
+++ b/mex/extractors/endnote/extract.py
@@ -67,7 +67,7 @@ def extract_endnote_records() -> list[EndnoteRecord]:
                     ),
                     publisher=find_text_from_record(record, "publisher/style"),
                     rec_number=find_text_from_record(record, "rec-number"),
-                    ref_type=ref_type.get("name")
+                    ref_type=ref_type.get("name")  # ref_type is stored differently
                     if isinstance(ref_type := record.find("ref-type"), ET.Element)
                     else None,
                     related_urls=findall_text_from_record(
--- a/mex/extractors/endnote/extract.py
+++ b/mex/extractors/endnote/extract.py
@@ -31,59 +31,60 @@ def extract_endnote_records() -> list[EndnoteRecord]:
     for file_name in file_names:
         response = connector.get_raw_file("endnote", file_name)
         file = defused_ET.fromstring(response.content)
-    for record in file.findall("*/record"):
-        if (
-            len(
-                authors := findall_text_from_record(
-                    record, "contributors/authors/author/style"
+
+        for record in file.findall("*/record"):
+            if (
+                len(
+                    authors := findall_text_from_record(
+                        record, "contributors/authors/author/style"
+                    )
+                )
+                >= cutoff_number_authors
+            ):
+                continue
+            results.append(
+                EndnoteRecord(
+                    database=find_text_from_record(record, "database"),
+                    abstract=find_text_from_record(record, "abstract/style"),
+                    authors=authors,
+                    call_num=find_text_from_record(record, "call-num/style"),
+                    custom3=find_text_from_record(record, "custom3/style"),
+                    custom4=find_text_from_record(record, "custom4/style"),
+                    custom6=find_text_from_record(record, "custom6/style"),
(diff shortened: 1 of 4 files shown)


mex/extractors/endnote/extract.py
```diff
@@ -1,12 +1,23 @@
...
+    """Finds all instances of items in path."""
+    return ["".join(node.itertext()) for node in record.findall(path)]
+
+
+def find_text_from_record(record: ET.Element, path: str) -> str | None:
+    """Finds the instance of an item in path."""
+    if isinstance(node := record.find(path), ET.Element):
+        return "".join(node.itertext())
+    return None
+
```
esinsj, line 20: I had to change Nics solution. For some strange reason "if node := record.find(path):" is False, even when node is assigned with a value. Eric's "isinstance" approached worked, so I took that.

mex/extractors/endnote/transform.py
```diff
@@ -318,6 +322,7 @@ def extract_endnote_bibliographic_resource(  # noqa: C901, PLR0915
                 )
                 load([created_org])
                 publisher.append(created_org.stableTargetId)
+                publisher_lookup_dict[publisher_string] = created_org.stableTargetId
```
esinsj: As the wikidata-helper returns an Identifier and loads the organization itself, I decided to use the dictionary for all entries, and have the self-created entries also loaded individually (and not collect all self-created orgs and load the values of the dictionary in a chunk at the end).

mex/extractors/endnote/extract.py
```diff
@@ -15,103 +26,65 @@ def extract_endnote_records() -> list[EndnoteRecord]:
     """
     connector = DropApiConnector.get()
     file_names = connector.list_files("endnote")
-    files: list[Any] = []
+    results: list[EndnoteRecord] = []
+    cutoff_number_authors = 50
```
erichesse: Could you put this into settings? That's cleaner for hardcoded values

mex/extractors/endnote/extract.py
```diff
@@ -15,103 +26,65 @@ def extract_endnote_records() -> list[EndnoteRecord]:
...
+                    periodical=findall_text_from_record(
+                        record, "periodical/full-title/style"
+                    ),
+                    pub_dates=findall_text_from_record(
+                        record, "dates/pub-dates/date/style"
+                    ),
+                    publisher=find_text_from_record(record, "publisher/style"),
+                    rec_number=find_text_from_record(record, "rec-number"),
+                    ref_type=ref_type.get("name")
+                    if isinstance(ref_type := record.find("ref-type"), ET.Element)
```
erichesse, line 72:
```suggestion
                    if isinstance(ref_type := find_text_from_record(record,"ref-type"), ET.Element)
```

mex/extractors/endnote/transform.py
```diff
@@ -306,10 +307,13 @@ def extract_endnote_bibliographic_resource(  # noqa: C901, PLR0915
         for publisher_string in [record.publisher, record.custom3]:
             if publisher_string is None:
                 continue
-            if publisher_org := get_wikidata_extracted_organization_id_by_name(
+            if publisher_string in publisher_lookup_dict:
```
erichesse: Does this even improve performance? Because it shouldn't as the wikidata calls are cached. We also said, we don't care about performance optimization of extractors?

mex/extractors/endnote/transform.py
erichesse: If anything load the created orgs at the end could help though, as it would ingest the created orgs in chunks

erichesse (commented):
Looks good. Few comments left

mex/extractors/endnote/transform.py
esinsj: Ok, This change is removed from the Branch now.

mex/extractors/endnote/extract.py
esinsj: done

erichesse (approved):

PR #29, opened 2025-07-10
feature/mx-1927 publishing with contact fallback

rababerladuseladim:
### Added

- add contact-point mini extractor to get default contact points

### Changes

- filter out unpublished persons from contact fields and set a fallback


mex/extractors/publisher/main.py
```diff
@@ -32,16 +43,20 @@ def publishable_merged_items() -> ItemsContainer[AnyMergedModel]:
...
+def publishable_persons() -> ItemsContainer[AnyMergedModel]:
+    """Get publishable persons from green-lit primary sources."""
     settings = Settings.get()
     connector = BackendApiConnector.get()
+    limit = 100
+    primary_sources = connector.fetch_extracted_items(
+        None, None, ["ExtractedPrimarySource"], 0, limit
+    )
+    if primary_sources.total > limit:
+        raise NotImplementedError
```
erichesse, line 55: Why is this necessary? If we assume `fetch_extracted_items` to work as intended, this should never occur

mex/extractors/publisher/transform.py
```diff
@@ -0,0 +1,26 @@
...
+from mex.extractors.publisher.fields import CONTACT_FIELDS_BY_CLASS_NAME
+from mex.extractors.utils import ensure_list
+
+
+def update_contact_where_needed(
+    item: AnyMergedModel,
+    allowed_contacts: Collection[AnyMergedIdentifier],
+    fallback_contact_identifiers: Iterable[AnyMergedIdentifier],
+) -> None:
+    """Update references in contact fields, where needed."""
```
erichesse, line 22: As the only `ContactPoint` Types are in the `contact` field, I would suggest hardcoding updating for the contact field in the YAGNI spirit

erichesse (commented):
Looks good, just two remarks

mex/extractors/publisher/main.py
cutoffthetop: `len(primary_sources.items)` should never be over 100, but the `total` still might be. the backend calculates the total separately from the "item retrieval" part of the request

mex/extractors/publisher/transform.py
cutoffthetop: true, fixed in a4286f6

erichesse (approved):

PR #35, opened 2025-07-15
feature/mx-1425-add_rki_org_as_unitOf

ZehraVictoria:
### Added

- ExtractedOrganizationalUnits have a link to the RKI ExtractedOrganization
### Changes

- use mex-common 0.63


tests/conftest.py
```diff
@@ -35,3 +35,13 @@ def extracted_organization_rki() -> ExtractedOrganization:
         hadPrimarySource=MergedPrimarySourceIdentifier.generate(123),
         officialName=["Robert Koch-Institut"],
     )
+
+
+@pytest.fixture
```
erichesse: I think `extracted_organization_rki` above should do the trick, so this fixture is not needed, right?

tests/pipeline/test_organigram.py
```diff
@@ -24,10 +25,13 @@ def extracted_primary_source_organigram() -> ExtractedPrimarySource:
 
 def test_extracted_organizational_units(
     extracted_primary_source_organigram: ExtractedPrimarySource,
+    extracted_organization_rki: ExtractedOrganization,
 ) -> None:
     units = cast(
         "list[ExtractedOrganizationalUnit]",
-        extracted_organizational_units(extracted_primary_source_organigram),
+        extracted_organizational_units(
```
erichesse, line 33: could you add an assertion for the setting of the field `unitOf` for one unit?

erichesse (commented):
Looks very good! Just two small issues

tests/pipeline/test_organigram.py
ZehraVictoria: fixed in https://github.com/robert-koch-institut/mex-extractors/pull/35/commits/0e6cdbb2cae4d51e15560876db10cfa83f4977a0

tests/conftest.py
ZehraVictoria: fixed in https://github.com/robert-koch-institut/mex-extractors/pull/35/commits/cb14a204d9b2b7f01830770c7d78ba6c5d390670 (a magical new fixture appeared, I don't remember `extracted_organization_rki` being there before)

erichesse (approved):

PR #37, opened 2025-07-15
feature/resolve dagster defs

cutoffthetop:
### PR Context

fixes this issue:

```
UserWarning: Found asset job named publisher of type <classedAssetJobDefinition'>
passed to `jobs` parameter. Starting in dagster 1.11, you must now use
Definitions.resolve_job_def to correctly retrieve this job definition.
```

### Changes

- ensure dagster definitions are resolved, for dagster 1.11 compat

### Added

- add mex-extractors version to asset and job metadata


mex/extractors/pipeline/base.py
```diff
@@ -147,12 +148,16 @@ def load_job_definitions() -> Definitions:
...
+    defs = Definitions(
         assets=assets,
         asset_checks=checks,
         jobs=jobs,
         resources=resources,
         schedules=schedules,
-        sensors=[monitor_jobs_sensor],
+        sensors=sensors,
     )
+    defs.get_repository_def().load_all_definitions()
```
erichesse: I am a little surprised, I can't find it in the docs. and the warning seems to suggest to use `resolve_job_def ` for the jobs.

erichesse (commented):
Back to you with one question

mex/extractors/pipeline/base.py
cutoffthetop: hm, true, i was thinking this would resolve all the jobs at once. but i guess it's not even needed here. it's enough if the resolving happens in `run_job_in_process`.

fixed it in 62079a4 (plus I added the package version to the metadata)

erichesse (dismissed):
looks good

PR #46, opened 2025-07-28
feature/mx-1927-remove-all-person-refs

cutoffthetop:
### Changes
- remove all broken person references in the publisher

### DIFF
--- a/mex/extractors/publisher/transform.py
+++ b/mex/extractors/publisher/transform.py
@@ -1,29 +1,42 @@
-from collections.abc import Collection, Iterable
+from collections.abc import Collection
 
-from mex.common.models import (
-    AnyMergedModel,
-    MergedAccessPlatform,
-    MergedActivity,
-    MergedPrimarySource,
-    MergedResource,
-)
+from mex.common.logging import logger
+from mex.common.models import AnyMergedModel
 from mex.common.types import AnyMergedIdentifier, MergedContactPointIdentifier
-
-MergedModelsWithContact = (
-    MergedAccessPlatform | MergedActivity | MergedPrimarySource | MergedResource
+from mex.extractors.publisher.fields import (
+    REFERENCED_ENTITY_TYPES_BY_FIELD_BY_CLASS_NAME,
 )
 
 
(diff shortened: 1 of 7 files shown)


mex/extractors/publisher/transform.py
```diff
@@ -1,29 +1,42 @@
...
+            if not identifiers and "MergedContactPoint" in ref_types:
+                identifiers = fallback_contact_identifiers
+            if not identifiers and item.model_fields[field].is_required():
+                logger.error(
+                    "%s(identifier='%s') has no valid references "
+                    "for required field %s, publishing broken references",
+                    item.entityType,
+                    item.identifier,
+                    field,
+                )
```
esinsj, line 44: item is still published without updates or changes. (?)

esinsj (commented):
two remarks:
- items with broken references still get published
- in extractor.py line 15 the primary source and entity type are not considered for getting the total = if there are only few items to extract the loop still runs over all items.

mex/extractors/publisher/transform.py
cutoffthetop: yep, because skipping the item could result in a recursive "de-publication" of items - which is less desirable than having a few broken connections, at the moment.
and if contact points are not allowed in that field, there is nothing to fallback on.

but you're right, i should update the CHANGELOG and remove the `all` from there. that's a lie.

cutoffthetop:
> in extractor.py line 15 the primary source and entity type are not considered for getting the total = if there are only few items to extract the loop still runs over all items.

good catch, fixed that in [1465806](https://github.com/robert-koch-institut/mex-extractors/pull/46/commits/1465806d87fcf77930c1258910cff2280ffab837)

mex/extractors/publisher/transform.py
cutoffthetop: updated changelog and docstring in [1465806](https://github.com/robert-koch-institut/mex-extractors/pull/46/commits/1465806d87fcf77930c1258910cff2280ffab837)

esinsj (approved):

PR #47, opened 2025-07-28
feature/update-mex-common-dependency-to-0.65

esinsj:
### Changes
- update mex-common to 0.65

as the Backend API Connector now allows to use a function to fetch ALL merged items, this is used where applicable. The according needed surrounding code (like first finding out the total number of items to be fetched), was therefore removed.

Also the tests needed to be adapted for using the new function and also for changing the positional arguments to keyword arguments int connector fetch functions.


erichesse (approved):
Looks good. Just a tiny issue. I would like `mocked_backend_api_connector` to be moved to tests/conftest.py or something reflecting that fact, that we want to use it more generally.

PR #48, opened 2025-07-29
Feature/mx 1856 send emails for informed consent

erichesse:
### Added

- finished consent-mailer pipeline


eckelmannf:
To run this locally u need https://github.com/mailhog/MailHog installed. 
I used the following args:
-hostname rki.de -ui-bind-addr 127.0.0.1:8025 -api-bind-addr 127.0.0.1:8025 -smtp-bind-addr 127.0.0.1:1025

tests/conftest.py
```diff
@@ -10,6 +10,7 @@
     "mex.common.testing.plugin",
     "tests.blueant.mocked_blueant",
     "tests.confluence_vvt.mocked_confluence_vvt",
+    "tests.consent_mailer.mocked_consent_backend_api_connector",
```
esinsj:
```suggestion
```
as the consent-mailer-job is ignored in mex/extractors/pipeline/base.py, it won't run in the tests/test_main.py, and therefore doesn't need to be imported here.

tests/consent_mailer/mocked_consent_backend_api_connector.py
esinsj, line 1: tldr: rename to "conftest.py"

For mocking we were told, to name locally-only used mocks (as is the case here) "conftest.py". These don't need to be imported outside the actual test folder. Only mocks which need to be used outside are called "mocked_XXYY" and get imported.

tests/consent_mailer/test_main.py
```diff
@@ -1,5 +1,75 @@
...
+MAIL_HOG_API_URL = "http://localhost:8025"
+
 
+@pytest.mark.usefixtures("mocked_consent_backend_api_connector")
 def test_job() -> None:
     assert run_job_in_process("consent_mailer")
+
+
+@pytest.mark.integration  # disabled on gh cli due to no mailHog available
+def test_send_consent_emails_to_persons() -> None:
```
esinsj: this test fails for me: assert 0 == 1

mex/extractors/consent_mailer/transform.py
```diff
@@ -0,0 +1,80 @@
...
+        person (MergedPerson): The person that needs to be transformed into a sendable
+        consent email.
+
+    Returns:
+        EmailMessage | None: The email message for the defined person or None if the
+        person doesn't have a @rki.de email address.
+    """
+    settings = Settings.get()
+
+    rki_emails = filter(lambda mail: mail.endswith("@rki.de"), person.email)
```
esinsj, line 33: could this filter also be moved to filter.py? I was a bit confused to find it here.

esinsj (commented):
Looks great. And thanks for the Mailhog hints. 
Some feedback in the comments.
Could you also merge main into the branch again? There are some conflicts with datenkompass-changes. In case you have questions, contact me.

tests/consent_mailer/test_main.py
eckelmannf: Yeah it was a bit tricky to configure Mailhog correctly. I think this test should pass, because its basicly the main test. We can try to set it up correctly together if u want to.

mex/extractors/consent_mailer/transform.py
eckelmannf: I was also thinking about the same thing. But since we designed the methods in the mob session and i only filled them, i was a bit careful about changing method signature or how they get called. How would u move it to another file? I think we should then redesign the whole method chain, because i need the persons to generate the email body and persons always consist of many emails (also the not rki related).

esinsj: Good question. maybe by creating a dictionary of rki-email and fullName in the filter function and use that for creating the email-body?

eckelmannf: As discussed in the daily. We keep the mail filtering as it is.

esinsj (dismissed):
Looks all good. But somehow the test for transform_person_to_sendable_email fails on Github (it passes locally), probably because the assets repo isn't merged yet!?
- could you update the branch with main again?

esinsj (dismissed):

mex/extractors/consent_mailer/main.py
```diff
@@ -11,8 +20,36 @@
...
+    all_ldap_persons_for_mailer: list[MergedPerson],
+) -> None:
+    """Send consent emails to the given persons.
+
+    Args:
+        all_ldap_persons_for_mailer (list[MergedPerson]): The list of persons that gets
+        the consent email.
+    """
+    settings = Settings.get()
+    [host, port] = settings.consent_mailer.smtp_server.split(":")
```
cutoffthetop:
```suggestion
    host, port = settings.consent_mailer.smtp_server.split(":")
```

mex/extractors/consent_mailer/main.py
```diff
@@ -11,8 +20,36 @@
...
+
+
+@asset(group_name="consent_mailer")
+def send_consent_emails_to_persons(
+    all_ldap_persons_for_mailer: list[MergedPerson],
+) -> None:
+    """Send consent emails to the given persons.
+
+    Args:
+        all_ldap_persons_for_mailer (list[MergedPerson]): The list of persons that gets
```
cutoffthetop:
```suggestion
        all_ldap_persons_for_mailer: The list of persons that gets
```

no need to duplicate the typing in the docstring

mex/extractors/consent_mailer/settings.py
```diff
@@ -0,0 +1,39 @@
...
+    )
+    mailpit_api_user: SecretStr = Field(
+        SecretStr("user"),
+        description="The username used for Basic HTTP auth against the mailpit api."
+        " USED FOR TESTS ONLY!",
+    )
+    mailpit_api_password: SecretStr = Field(
+        SecretStr("password"),
+        description="The password used for Basic HTTP auth against the mailpit api."
+        "USED FOR TESTS ONLY!",
```
cutoffthetop:
```suggestion
        " USED FOR TESTS ONLY!",
```

mex/extractors/consent_mailer/settings.py
```diff
@@ -0,0 +1,39 @@
...
+    )
+    schedule: str = Field(
+        "0 0 * * *",
```
cutoffthetop: let's use a safer default, so we don't accidentally launch a daily spam bot

```suggestion
        "0 0 30 2",  # Defaults to Feb 30, aka "never"
```

tests/consent_mailer/test_main.py
```diff
@@ -1,5 +1,115 @@
...
+    )
+
+
+def _get_messages() -> Any:  # noqa: ANN401
+    settings = Settings.get()
+    return requests.get(
+        f"{settings.consent_mailer.mailpit_api_url}/api/v1/messages",
+        timeout=3,
+        verify=_req_verify(),
+        headers=_req_header() | {"accept": "application/json"},
```
cutoffthetop:
```suggestion
        auth=(
            settings.consent_mailer.mailpit_api_user.get_secret_value(),
            settings.consent_mailer.mailpit_api_password.get_secret_value()
        ),
```

accept header is not really needed, and requests can do basic-auth out of the box with a (user, pw) tuple

tests/consent_mailer/test_main.py
```diff
@@ -1,5 +1,115 @@
...
+    ).json()
+
+
+@pytest.mark.integration  # disabled on gh cli due to no mailHog available
```
cutoffthetop:
```suggestion
@pytest.mark.integration  # disabled on gh cli due to missing mailpit, stopgap MX-1993
```

tests/consent_mailer/test_main.py
```diff
@@ -1,5 +1,115 @@
...
+
+    envelop = _get_messages()
+    assert envelop["total"] == 1
+    assert all(
+        all(x["Address"].endswith("rki.de") for x in message["To"])
+        for message in envelop["messages"]
+    )
+
+
+@pytest.mark.integration  # disabled on gh cli due to no mailHog available
```
cutoffthetop:
```suggestion
@pytest.mark.integration  # disabled on gh cli due to missing mailpit, stopgap MX-1993
```

tests/consent_mailer/test_main.py
```diff
@@ -1,5 +1,115 @@
...
+def _req_header() -> dict[str, str]:
+    settings = Settings.get()
+    return {
+        "Authorization": "Basic "
+        + base64.b64encode(
+            f"{settings.consent_mailer.mailpit_api_user.get_secret_value()}:{settings.consent_mailer.mailpit_api_password.get_secret_value()}".encode()
+        ).decode()
+    }
+
+
```
cutoffthetop:
```suggestion
```

see below

cutoffthetop (changes requested):
looks good!

a few simplifying suggestions from me. most importantly though, i'd like a safer default for the cron. and a quick main-merge is needed.

after that, ready to merge 🚀

mex/extractors/consent_mailer/settings.py
eckelmannf: "0 0 30 2" is not valid. expecting u mean "0 0 30 2 *"

eckelmannf: Seems dagster is validating this aswell. I change it to a far future 🤖

eckelmannf: since dagster is validating the schedule and 30.2 is invalid i have to change it to "0 0 29 2 *". Any other ideas?

cutoffthetop: dagster uses croniter under the hood and that does allow for a seventh slot to set a year
https://github.com/pallets-eco/croniter?tab=readme-ov-file#about-year

-> `0 0 1 1 * 0 1970` should work as a sane default that will never happen (at least not for the next 9945 years)

eckelmannf: Seems like dagster doing some extra validation.
`
Found invalid cron schedule '0 0 1 1 * 0 1970' for schedule 'None''. Dagster recognizes standard cron expressions consisting of 5 fields.
`

esinsj: it is also possible to not set a schedule at all. This is how I did it for Datenkompass. If that is an option?

eckelmannf: How did u achieve this? When I set it to an empty string, dagster also complains :(
`Found invalid cron schedule '' for schedule 'None''. Dagster recognizes standard cron expressions consisting of 5 fields.`

mex/extractors/consent_mailer/settings.py
```diff
@@ -0,0 +1,39 @@
...
+        description="The username used for Basic HTTP auth against the mailpit api."
+        " USED FOR TESTS ONLY!",
+    )
+    mailpit_api_password: SecretStr = Field(
+        SecretStr("password"),
+        description="The password used for Basic HTTP auth against the mailpit api."
+        " USED FOR TESTS ONLY!",
+    )
+    schedule: str = Field(
+        "0 0 29 2 *",
```
eckelmannf: Used Feb 29 since i couldnt get anything else ("", None) working.

mex/extractors/consent_mailer/settings.py
```diff
@@ -21,8 +21,8 @@ class ConsentMailerSettings(BaseSettings):
         description="The password used for Basic HTTP auth against the mailpit api."
         " USED FOR TESTS ONLY!",
     )
-    schedule: str = Field(
-        "0 0 29 2 *",
+    schedule: str | None = Field(
```
eckelmannf, line 24: Used None now as suggested by @esinsj

esinsj (approved):
Nice. I meant to set no Scheduler at all. But I like the solution to set the scheduler to None even better.

cutoffthetop (approved):

PR #53, opened 2025-08-11
Feature/mx 1944 add datenkompass resources

esinsj:
### Added
- more activities added to datenkompass


esinsj:
outdated. 
Replaces by new PR

PR #54, opened 2025-08-14
Feature/mx 1976 fix journal titles endnote

esinsj:
### Changes
- update extraction of journal for Bib resource
- update loading of publisher organization for Bib resource


erichesse (approved):
Great! Looks good

PR #55, opened 2025-08-18
Feature/improvements for mex common dependency to 0.65

esinsj:
### Changes
- improve usage of backend endpoints for merged items

The endpoint usage needed to be adapted for datenkompass to accept more reference fields than primary source (see later PR). In that scope I also generalized the endpoint usage for publisher. And I threw out the preview item endpoint, as we don't expect changes for the merged item endpoint anymore and can keep using it. I combined all those changes into this PR.


erichesse (dismissed):
Looks good!

erichesse (approved):

PR #61, opened 2025-08-26
Feature/mx 1894 igs variable group variable mandatory

erichesse:
### Added

- extract mandatory fields for igs variables and variable groups


mex/extractors/igs/transform.py
```diff
@@ -51,14 +67,14 @@ def transform_igs_schemas_to_resources(
...
-        pathogen: ExtractedResource(
+    return [
+        ExtractedResource(
             accessRestriction=igs_resource_mapping.accessRestriction[0]
             .mappingRules[0]
             .setValues,
             contact=contact,
             hadPrimarySource=extracted_primary_source_igs.stableTargetId,
-            identifierInPrimarySource=f"IGS_{pathogen}",
+            identifierInPrimarySource=f"IGS_{igs_info.title}_v{igs_info.version}",
```
esinsj, line 69: not sure, but as this is needed three times, and is the identifierInPrimarySource, would it make sense to move this into a separate function, to be sure it's always created the same way?

mex/extractors/igs/filter.py
```diff
@@ -0,0 +1,21 @@
...
+        igs_schemas: igs schemas
+
+    Returns:
+        filter creation schemas
+    """
+    return {
+        name: schema
+        for name, schema in igs_schemas.items()
+        if (isinstance(schema, IGSPropertiesSchema))
+        and "Creation" in name
```
esinsj:
```suggestion
        and name.endswith("Creation")
```

mex/extractors/igs/transform.py
```diff
@@ -51,14 +67,14 @@ def transform_igs_schemas_to_resources(
...
             theme=igs_resource_mapping.theme[0].mappingRules[0].setValues,
             title=title_by_pathogen.get(
```
esinsj: the title extraction needs to be updated

mex/extractors/igs/transform.py
```diff
@@ -113,3 +129,139 @@ def transform_igs_access_platform(
...
+        list of ExtractedVariable
+    """
+    enums_by_property_name = get_enums_by_property_name(igs_schemas)
+    extracted_variables: list[ExtractedVariable] = []
+    resource_identifier = f"IGS_{igs_info.title}_v{igs_info.version}"
+    for schema_name, schema in igs_schemas.items():
+        if not isinstance(schema, IGSPropertiesSchema):
+            continue
+        for property_name, schema_property in schema.properties.items():
+            belongs_to = (
```
esinsj, line 262: Something doesn't work properly here, because there are ExtractedVariables with empty "belongsTo", which shouldn't be possible

mex/extractors/igs/transform.py
```diff
@@ -113,3 +129,139 @@ def transform_igs_access_platform(
...
+                resource_identifier
+            ]
+            value_set = enums_by_property_name.get(property_name, [])
+            extracted_variables.append(
+                ExtractedVariable(
+                    belongsTo=belongs_to,
+                    dataType=data_type,
+                    description=schema_property.get("title"),
+                    hadPrimarySource=extracted_primary_source_igs.stableTargetId,
+                    identifierInPrimarySource=property_name,
```
esinsj, line 284: there are properties with the same name, like "sample", which are at the moment extracted separately and not getting combined with all "belongsTo" collected.

mex/extractors/igs/transform.py
```diff
@@ -113,3 +129,139 @@ def transform_igs_access_platform(
...
+        igs_info: igs info
+
+    Returns:
+        list of ExtractedVariable
+    """
+    enums_by_property_name = get_enums_by_property_name(igs_schemas)
+    extracted_variables: list[ExtractedVariable] = []
+    resource_identifier = f"IGS_{igs_info.title}_v{igs_info.version}"
+    for schema_name, schema in igs_schemas.items():
+        if not isinstance(schema, IGSPropertiesSchema):
```
esinsj, line 259: the pathogenes get skipped and don't get extracted with this clause.

mex/extractors/igs/transform.py
```diff
@@ -51,14 +67,14 @@ def transform_igs_schemas_to_resources(
...
+    return [
+        ExtractedResource(
             accessRestriction=igs_resource_mapping.accessRestriction[0]
             .mappingRules[0]
             .setValues,
             contact=contact,
             hadPrimarySource=extracted_primary_source_igs.stableTargetId,
-            identifierInPrimarySource=f"IGS_{pathogen}",
+            identifierInPrimarySource=f"IGS_{igs_info.title}_v{igs_info.version}",
             theme=igs_resource_mapping.theme[0].mappingRules[0].setValues,
```
esinsj, line 70: keywords, language, meshId, ... are missing

mex/extractors/igs/transform.py
```diff
@@ -113,3 +129,139 @@ def transform_igs_access_platform(
...
+        )
+        extracted_variable_groups.append(
+            ExtractedVariableGroup(
+                containedBy=contained_by,
+                hadPrimarySource=extracted_primary_source_igs.stableTargetId,
+                identifierInPrimarySource=schema_name,
+                label=label,
+            )
+        )
+    return extracted_variable_groups
```
esinsj, line 157: the "Pathogen" variable group is missing

esinsj (changes requested):
some changes needed

mex/extractors/igs/transform.py
erichesse: I think it's easier to understand an f string, which is very straigt forward to understand than hiding it behind a function, which you need to understand.

mex/extractors/igs/transform.py
erichesse: see above

mex/extractors/igs/transform.py
erichesse: Why not?
```
belongsTo:
  - fieldInPrimarySource: "*Creation"
    locationInPrimarySource: components/schemas/
    mappingRules:
      - rule: Match with identifierInPrimarySource of the variable-group.
          Assign stableTargetId of respective item described by /mappings/../igs/variable-group.
```
The empty cases I identified, don't match the *Creation pattern

mex/extractors/igs/transform.py
erichesse: I made a comment in the wrap up ticket

mex/extractors/igs/transform.py
erichesse: https://github.com/robert-koch-institut/mex-extractors/pull/61/commits/5e817b4f06b443c4c478ffb8fb85438888217d8d

mex/extractors/igs/transform.py
erichesse: It's supposed to only extract the mandatory fields

mex/extractors/igs/transform.py
erichesse: https://github.com/robert-koch-institut/mex-extractors/pull/61/commits/5e817b4f06b443c4c478ffb8fb85438888217d8d

mex/extractors/igs/transform.py
esinsj: I think a clear function name like "create_identifier_in_primary_source" can also be easily understood and is less error prone, e.g. if the identifierInPrimarySource have to be changed for some reason. but no hard veto here.

mex/extractors/igs/transform.py
esinsj: > see above

Just to be sure, do you mean "It's supposed to only extract the mandatory fields"? (The order of comments is different in "Conversations" Tab and "Files changed" Tab.)

mex/extractors/igs/transform.py
esinsj: Ah, because of the statement "Only create variable-groups for schemas that have "Creation" as suffix." in the variable-group mapping, I found it sensible that only variables should be extracted, which belong to that variable groups. This should be clarified by mdmgmt.

mex/extractors/igs/transform.py
esinsj: Saw it. Not sure, if it's a problem of identifierInPrimarySource (I thought that, too) or if it really just is the same variable (that was my next idea). Probably a point for mdmgmt.

mex/extractors/igs/transform.py
esinsj: Ah, I see. Thanks for the reminder.

mex/extractors/igs/transform.py
```diff
@@ -113,3 +129,168 @@ def transform_igs_access_platform(
...
+        f"IGS_{igs_info.title}_v{igs_info.version}"
+    ]
+    for schema_name, schema in igs_schemas.items():
+        data_type: str | None = None
+        if schema_name == "Pathogen" and isinstance(schema, IGSEnumSchema):
+            for enum in schema.enum:
+                belongs_to = (
+                    [extracted_igs_variable_group_ids_by_igs_identifier[schema_name]]
+                    if schema_name in extracted_igs_variable_group_ids_by_igs_identifier
+                    else []
```
esinsj, line 242: Wouldn't it be better to let the extractor fail if the group is missing? Maybe mdmgmt has an opinion about that?

mex/extractors/igs/transform.py
```diff
@@ -113,3 +129,168 @@ def transform_igs_access_platform(
...
+                )
+                data_type = schema.type
```
esinsj: "data_type" assignment could be moved above the "for enum" loop.

esinsj (commented):
Thanks for the clarifications.
The code looks really good to me. Two small ideas for the pathogen code.
Would it be worth adding some tests for the transformation functions?

mex/extractors/igs/transform.py
erichesse: yes

mex/extractors/igs/transform.py
erichesse: I don't understand what should be clarified

mex/extractors/igs/transform.py
erichesse: Added a comment in the wrap up ticket

mex/extractors/igs/transform.py
erichesse: resource.title needs to be updated

mex/extractors/igs/transform.py
erichesse: https://github.com/robert-koch-institut/mex-extractors/pull/61/commits/88405171beb1a53e3ccdf52ee0cfa8df3b0b4ebb

mex/extractors/igs/transform.py
erichesse: https://github.com/robert-koch-institut/mex-extractors/pull/61/commits/88405171beb1a53e3ccdf52ee0cfa8df3b0b4ebb

mex/extractors/igs/transform.py
esinsj: Commented in WrapUp

mex/extractors/igs/transform.py
esinsj: You probably mean [d0c5403](https://github.com/robert-koch-institut/mex-extractors/pull/61/commits/d0c5403da748fedeb8654a4f7b0ead5dda259525)
The solution there doesn't give the title specified in the mapping, but an ExtractedResource validation error

esinsj (changes requested):
What do you think about my comment on tests?

mex/extractors/igs/transform.py
erichesse: Oh, I finally got it:
https://github.com/robert-koch-institut/mex-extractors/pull/61/commits/203e470d269d69de8ca44170c81187a762749df1

mex/extractors/igs/transform.py
```diff
@@ -39,38 +54,28 @@ def transform_igs_schemas_to_resources(
...
-        for rule in igs_resource_mapping.title[0].mappingRules
-        if rule.setValues and rule.forValues
-    }
+    title = igs_resource_mapping.title[0].mappingRules[0].setValues
     unit_in_charge = (
         unit_stable_target_ids_by_synonym[for_value[0]]
         if (for_value := igs_resource_mapping.unitInCharge[0].mappingRules[0].forValues)
         else []
     )
     pathogen_schema = cast("IGSEnumSchema", igs_schemas["Pathogen"])
```
esinsj:
```suggestion
```
not needed, or?

mex/extractors/igs/transform.py
```diff
@@ -39,38 +54,28 @@ def transform_igs_schemas_to_resources(
...
-                pathogen,
-                Text(
-                    value=f"Integrierte Genomische Surveillance {pathogen}",
-                    language=TextLanguage.DE,
-                ),
-            ),
+            title=title,
             unitInCharge=unit_in_charge,
         )
         for pathogen in pathogen_schema.enum
```
esinsj:
```suggestion
```

obsolete (same as above)

mex/extractors/igs/transform.py
```diff
@@ -113,3 +118,177 @@ def transform_igs_access_platform(
...
+        data_type: str | None = None
+        if schema_name == "Pathogen" and isinstance(schema, IGSEnumSchema):
+            data_type = schema.type
+            for enum in schema.enum:
+                belongs_to = (
+                    [extracted_igs_variable_group_ids_by_igs_identifier[schema_name]]
+                    if schema_name in extracted_igs_variable_group_ids_by_igs_identifier
+                    else []
+                )
+                description = description_by_enum.get(enum)
```
esinsj, line 244: why this change? as description = title, wouldn't it make sense to fail if there's no title? That should fail and be handled, instead of missed, or?

mex/extractors/igs/transform.py
```diff
@@ -113,3 +118,177 @@ def transform_igs_access_platform(
...
+        property_name: enum_by_schema_name[schema_name]
+        for schema in igs_schemas.values()
+        if isinstance(schema, IGSPropertiesSchema)
+        for property_name, properties in schema.properties.items()
+        if (
+            "$ref" in properties
+            and (schema_name := properties["$ref"].split("/")[-1])
+            in enum_by_schema_name
+        )
+        or (
```
esinsj, line 186: <del>I don't believe that you need this or-clause, as I think you only need variables of group *creation and pathogen. And they don't have this case.</del> (incorrect)


Some enums are missing in the dict, like the second example of the exampleInPrimarySource for valueSet in the variable mapping.

esinsj (commented):
Some comments.
For some reason now the test_connector.py fails. And I didn't notice before that the test function naming in that file has a hickup ;)

mex/extractors/igs/transform.py
erichesse: Which title do you mean?

mex/extractors/igs/transform.py
erichesse: Without the or clause the extractor fails on me

erichesse: I made a comment in the wrap up ticket for 2)

erichesse:
> Some comments. For some reason now the test_connector.py fails. And I didn't notice before that the test function naming in that file has a hickup ;)

should be fixed now: https://github.com/robert-koch-institut/mex-extractors/pull/61/commits/0cdbb4a1fd88eefea57a47ccdf5dcbbd16e35be0

mex/extractors/igs/transform.py
esinsj: > Without the or clause the extractor fails on me

You're absolutely right, I was wrong. Forget my comment.

mex/extractors/igs/transform.py
esinsj: I'm sorry, I wrote the wrong thing. I meant:
why this change? as description = lookup in mapping, wouldn't it make sense to fail if there's no mapping case? Maybe mdmgmt has an opinion about it.

mex/extractors/igs/transform.py
```diff
@@ -6,23 +6,36 @@
...
+    IGSEnumSchema,
+    IGSInfo,
+    IGSPropertiesSchema,
+    IGSSchema,
 )
-from mex.common.types import MergedOrganizationalUnitIdentifier, Text, TextLanguage
-from mex.extractors.igs.model import IGSEnumSchema, IGSSchema
 
 
 def transform_igs_schemas_to_resources(
```
esinsj:
```suggestion
def transform_igs_info_to_resource(
```
would it make sense to rename the function ten?

mex/extractors/igs/main.py
```diff
@@ -85,21 +99,26 @@ def extracted_igs_contact_points_by_mail(
...
     extracted_primary_source_igs: ExtractedPrimarySource,
     igs_resource_mapping: dict[str, Any],
     extracted_igs_contact_points_by_mail: dict[str, ExtractedContactPoint],
     unit_stable_target_ids_by_synonym: dict[str, MergedOrganizationalUnitIdentifier],
-) -> dict[str, ExtractedResource]:
+) -> dict[str, MergedResourceIdentifier]:
     """Transform IGS resource from IGS schemas."""
-    return transform_igs_schemas_to_resources(
-        igs_schemas,
+    extracted_resources = transform_igs_schemas_to_resources(
```
esinsj:
```suggestion
    extracted_resources = transform_igs_info_to_resource(
```

tests/igs/test_connector.py
```diff
@@ -4,10 +4,10 @@
 
 
 @pytest.mark.integration
-def test_datscha_connector_item_urls() -> None:
+def test_igs_connector_item_urls() -> None:
```
esinsj:
```suggestion
def test_igs_connector_json_from_api() -> None:
```

pdm.lock
esinsj, line 1: isn't a general dependency update something that should have its own PR?

esinsj (commented):
Looks very good so far. Three more ideas / questions.

mex/extractors/igs/transform.py
erichesse: made a comment in the wrap up ticket

pdm.lock
erichesse: I think it doesn't matter too much

mex/extractors/igs/main.py
erichesse: https://github.com/robert-koch-institut/mex-extractors/pull/61/commits/de23aff08c43480f87328218e87329f0b1f4e544

mex/extractors/igs/transform.py
erichesse: https://github.com/robert-koch-institut/mex-extractors/pull/61/commits/de23aff08c43480f87328218e87329f0b1f4e544

tests/igs/test_connector.py
erichesse: https://github.com/robert-koch-institut/mex-extractors/pull/61/commits/6e713cd3d94d63d956bd9893b5c6da8b80947288

mex/extractors/igs/main.py
erichesse: https://github.com/robert-koch-institut/mex-extractors/pull/61/commits/6e713cd3d94d63d956bd9893b5c6da8b80947288

esinsj (approved):

PR #62, opened 2025-08-26
feature/return-loaded-assets

cutoffthetop:
### PR Context

I wanted to debug one of the assets that just had a "load" statement in the last line, but no return. The lack of a return statement made it impossible to inspect the asset in the dagster storage, because it was not materialized to disk. This PR adds return values to all assets to make debugging easier in the future.

### Added

- add return values to all assets for debugging and bookkeeping

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- add return values to all assets for debugging and bookkeeping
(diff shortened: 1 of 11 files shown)


erichesse (approved):
Looks good, just one question: we dont care about `publish_to_s3`?

cutoffthetop:
> Looks good, just one question: we dont care about `publish_to_s3`?

no, i thought it wouldn't matter for those, since they already have a _papertrail_ (on s3)

PR #75, opened 2025-09-25
Feature/update mex common dependency 1.4

erichesse:
**Note**: automatic language detection will be kept for now as discussed with mdmtmg

### Added

- update mex-common dependency to 1.4


rababerladuseladim (dismissed):
The test output contains a huuuuuuuge number of deprecation warnings for ldap. Do you want to fix these as part of this PR or rather separately?

See https://github.com/robert-koch-institut/mex-extractors/actions/runs/18001489452/job/51211611543

rababerladuseladim (approved):

PR #84, opened 2025-10-08
Feature/mx 2048 dagster job to purge old dagster runs

esinsj:
### Added
- purge scripts to clean up old dagster runs and files


mex/extractors/system/main.py
```diff
@@ -0,0 +1,86 @@
...
+    settings = Settings.get()
+    # Define the time threshold
+    cutoff_date = datetime.datetime.now(tz=UTC) - datetime.timedelta(
+        days=settings.system.max_run_age_in_days
+    )
+
+    collected_old_run_records: list[str] = []
+    cursor = None
+
+    while True:
```
erichesse:
```suggestion

```
This should basically never be used, especially if terminal conditions are not certain to be reached. If something goes wrong this run keeps running.

I would suggest to fetch enough runs, that the number obviously surpasses the number of daily runs. That might not catch all runs in one go, but will chip away some every day and keeps this maintenance job slim

mex/extractors/system/main.py
```diff
@@ -0,0 +1,86 @@
...
+        # Update cursor to continue pagination
+        cursor = old_run_records[-1].dagster_run.run_id
+
+    return collected_old_run_records
+
+
+@asset(group_name="system_clean_up")
+def system_clean_up_dagster_files(
+    system_fetch_old_dagster_run_ids: list[str],
+) -> list[str]:
```
erichesse, line 34: Why do we need a return value here?

mex/extractors/system/main.py
```diff
@@ -0,0 +1,86 @@
...
+        len(deleted_file_ids),
+        storage_path,
+    )
+    return deleted_file_ids
+
+
+@asset(group_name="system_clean_up", deps=[AssetKey("system_clean_up_dagster_files")])
+def system_clean_up_dagster_runs(
+    system_fetch_old_dagster_run_ids: list[str],
+) -> list[str]:
```
erichesse, line 61: Why do we need a return value here?

erichesse (commented):
Looks good so far.

But please no `while True`

mex/extractors/system/main.py
esinsj: you're absolutely right! Thanks for pointing that out. 
I changed it back to the initial approach with batches of 100. There are ~20 runs/day at the moment. If the cleanup script runs daily, that should be fine.
[2349dda](https://github.com/robert-koch-institut/mex-extractors/pull/84/commits/2349dda47cd028def52d77b5ff8f1d0c3da3af77)

mex/extractors/system/main.py
esinsj: We don't. I copied it from other extractors which give a return value which isn't used further, like https://github.com/robert-koch-institut/mex-extractors/blob/main/mex/extractors/biospecimen/main.py#L97. I was wondering myself, but it made testing easier, so I found it handy 😅

mex/extractors/system/main.py
esinsj: see other comment.

erichesse (approved):

PR #91, opened 2025-10-23
Feature/mx 1698 use primary source id convenience function

erichesse:
### Changes
- use primary source id convenience function


mex/extractors/wikidata/helpers.py
```diff
@@ -28,7 +28,7 @@ def get_wikidata_organization_by_id(wikidata_id: str) -> ExtractedOrganization |
         msg = "Primary source for wikidata not found"
         raise MExError(msg)
     extracted_organization = transform_wikidata_organization_to_extracted_organization(
-        wikidata_organization, wikidata_primary_source
+        wikidata_organization, wikidata_primary_source.stableTargetId
```
esinsj: just as an idea:
```suggestion
        wikidata_organization, get_extracted_primary_source_id_by_name("wikidata")
```
and delete lines 26-29

esinsj (dismissed):
wow, what a huge and complex change. 🤯
I would have opted for loading the identifier once per extractor and then pass it to the using methods (like done with the assets primary Source before). But I'm fine with it.

esinsj (approved):

PR #92, opened 2025-10-23
Feature/mx 2014 opendata update according to resource mapping

esinsj:
### Changes
- update open data resource according to mapping changes

#### Further information:
* This PR only address the changes of the Resource mapping. the Other mapping changes will be addressed in different tickets.
* the unitInCharge topic was solved with a global constant. There is ticket MX-2075 to find a clean solution (as discussed with ND)

### DIFF
--- a/mex/extractors/open_data/transform.py
+++ b/mex/extractors/open_data/transform.py
@@ -351,10 +351,13 @@ def transform_open_data_parent_resource_to_mex_resource(  # noqa: PLR0913
             in resource_mapping.license[0].mappingRules[0].forValues
             else None
         )
-        if resource_mapping.contributingUnit[0].mappingRules[0].forValues:
-            contributing_unit = unit_stable_target_ids_by_synonym.get(
+        contributing_unit = (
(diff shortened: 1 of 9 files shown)


mex/extractors/open_data/transform.py
```diff
@@ -305,6 +310,8 @@ def transform_open_data_parent_resource_to_mex_resource(  # noqa: PLR0913
                 for unit_id in unit_list
             }
         )
+        if not unit_in_charge:
+            unit_in_charge = [unit_stable_target_ids_by_synonym[FALLBACK_UNIT]]
```
erichesse, line 314: there seems to be yet another mapping rule using `metadata > creators > name`

erichesse (commented):
Looks good, it just seems, one mapping rule is missing

mex/extractors/open_data/transform.py
esinsj: yes, this is addressed in line 304, or do you mean something different?

erichesse (approved):

PR #93, opened 2025-10-28
Feature/mx 2078 igs mapping update

erichesse:
### Changes

- implement igs resource mapping update

### DIFF
--- a/mex/extractors/igs/transform.py
+++ b/mex/extractors/igs/transform.py
@@ -270,8 +270,7 @@ def transform_igs_schemas_to_variables(
 
     Args:
         igs_schemas: igs schemas by schema name
-        igs_extracted_resource_id:
-            igs resources by pathogen
+        igs_extracted_resource_id: stableTargetId of the igs resource
         extracted_igs_variable_group_ids_by_igs_identifier:
             igs variable group by identifier in primary source
         variable_mapping: variable mapping default values
--- a/mex/extractors/igs/transform.py
+++ b/mex/extractors/igs/transform.py
@@ -191,8 +191,7 @@ def transformed_igs_schemas_to_variable_group(
 
     Args:
         filtered_schemas: filtered IGS schemas
-        igs_extracted_resource_id:
-            dict of igs resource ids
+        igs_extracted_resource_id: stableTargetId of the igs resource
 
 
     Returns:
--- a/mex/extractors/igs/transform.py
+++ b/mex/extractors/igs/transform.py
@@ -19,7 +19,6 @@ from mex.common.types import (
 )
 from mex.extractors.igs.model import (
     IGSEnumSchema,
-    IGSInfo,
     IGSPropertiesSchema,
     IGSSchema,
 )
@@ -28,18 +27,16 @@ from mex.extractors.primary_source.helpers import (
 )
 
 
-def transform_igs_info_to_resources(  # noqa: PLR0913
-    igs_info: IGSInfo,
+def transform_igs_extracted_resource(
     igs_resource_mapping: ResourceMapping,
     igs_extracted_contact_points_by_mail_str: dict[str, ExtractedContactPoint],
     unit_stable_target_ids_by_synonym: dict[str, MergedOrganizationalUnitIdentifier],
     igs_extracted_access_platform: ExtractedAccessPlatform,
     extracted_organization_rki: ExtractedOrganization,
-) -> list[ExtractedResource]:
+) -> ExtractedResource:
     """Transform IGS schemas to extracted resources.
 
     Args:
-        igs_info: igs info
         igs_resource_mapping: IGS resource mapping
         igs_extracted_contact_points_by_mail_str: extracted IGS contact points by mail
         unit_stable_target_ids_by_synonym: merged organizational units by name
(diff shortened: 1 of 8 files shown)


mex/extractors/igs/transform.py
```diff
@@ -192,27 +185,22 @@ def transform_igs_access_platform(
...
-    igs_info: IGSInfo,
+    igs_extracted_resource_id: MergedResourceIdentifier,
 ) -> list[ExtractedVariableGroup]:
     """Transform IGS schema to extracted variable group.
 
     Args:
         filtered_schemas: filtered IGS schemas
-        igs_extracted_resource_ids_by_identifier_in_primary_source:
+        igs_extracted_resource_id:
             dict of igs resource ids
```
rababerladuseladim:
```suggestion
        igs_extracted_resource_id: stableTargetId of the igs resource
```

mex/extractors/igs/transform.py
```diff
@@ -270,29 +258,25 @@ def get_enums_by_property_name(
...
     variable_pathogen_mapping: VariableMapping,
-    igs_info: IGSInfo,
 ) -> list[ExtractedVariable]:
     """Transform igs schemas to variables.
 
     Args:
         igs_schemas: igs schemas by schema name
-        igs_extracted_resource_ids_by_identifier_in_primary_source:
+        igs_extracted_resource_id:
             igs resources by pathogen
```
rababerladuseladim:
```suggestion
        igs_extracted_resource_id: stableTargetId of the igs resource
```

tests/igs/test_transform.py
```diff
@@ -93,12 +92,9 @@ def test_transform_igs_access_platform(
...
     igs_schemas: dict[str, IGSSchema],
-    igs_info: IGSInfo,
 ) -> None:
     extracted_variable_groups = transformed_igs_schemas_to_variable_group(
-        igs_schemas,
-        {"IGS_test_title_vtest_version": MergedResourceIdentifier.generate(seed=42)},
-        igs_info,
+        igs_schemas, MergedResourceIdentifier.generate(seed=42)
     )
     expected = {
```
rababerladuseladim, line 99: Could you put `MergedResourceIdentifier.generate(seed=42)` into a variable, e.g. `resource_identifier` and reuse that in the expected content? Preferably also for the `hadPrimarySource`.

```
    expected = {
        "hadPrimarySource": "cT4pY9osJlUwPx5ODOGLvk",
        "identifierInPrimarySource": "Pathogen",
        "containedBy": [resource_identifier],
        "label": [{"value": "Pathogen", "language": "en"}],
        "identifier": Joker(),
        "stableTargetId": Joker(),
}
```

tests/igs/test_transform.py
```diff
@@ -123,15 +119,13 @@ def test_transform_igs_schemas_to_variables(
     igs_schemas: dict[str, IGSSchema],
     igs_variable_mapping: VariableMapping,
     igs_variable_pathogen_mapping: VariableMapping,
-    igs_info: IGSInfo,
 ) -> None:
     extracted_variables = transform_igs_schemas_to_variables(
         igs_schemas,
-        {"IGS_test_title_vtest_version": MergedResourceIdentifier.generate(seed=42)},
+        MergedResourceIdentifier.generate(seed=42),
         {"Pathogen": MergedVariableGroupIdentifier.generate(seed=43)},
```
rababerladuseladim, line 126: Same as for variable groups: could you move `MergedResourceIdentifier.generate(seed=42)` and `MergedResourceIdentifier.generate(seed=43)` into variables and use them in the expected content?

rababerladuseladim (changes requested):
Neat!

Just some tiny remarks.

tests/igs/test_transform.py
erichesse: My comments didn't get send. I said I didn't get the point as it creates more overhead and makes the test more complicated to read.

tests/igs/test_transform.py
erichesse: same here

tests/igs/test_transform.py
rababerladuseladim: The point is to make sure that the correct identifier is used in the field `hadPrimarySource`. In the current state, readers have to know that `MergedResourceIdentifier.generate(seed=42)` translates into `cT4pY9osJlUwPx5ODOGLvk` for checking that the correct identifier is used. Hence, I'd argue that it makes this test more readable and easier to understand.

rababerladuseladim: To be explicit, I suggest to change:

```python
    extracted_variable_groups = transformed_igs_schemas_to_variable_group(
        igs_schemas, MergedResourceIdentifier.generate(seed=42)
    )
    expected = {
        "hadPrimarySource": "cT4pY9osJlUwPx5ODOGLvk",
        {...}
    }
```

into

```python
    resource_identifier = MergedResourceIdentifier.generate(seed=42)
    extracted_variable_groups = transformed_igs_schemas_to_variable_group(
        igs_schemas, resource_identifier 
    )
    expected = {
        "hadPrimarySource": resource_identifier,
        {...}
    }

```

erichesse: I think this is obfuscating things by hiding them behind a variable. And also not correct, as there should be a MergedPrimarySourceIdentifier here, which is also already validated as such by the pydantic model.

rababerladuseladim: Well, I strongly disagree and rather think that `cT4pY9osJlUwPx5ODOGLvk` obfuscates that this is the `resource_identifier`. But I give in

rababerladuseladim (approved):

PR #96, opened 2025-10-31
feature/mx-1661 organigram convenience function

erichesse:
### PR Context
- follow-up to https://github.com/robert-koch-institut/mex-common/pull/566

### Added

- add convenience function to get units by synonym


mex/extractors/organigram/helpers.py
```diff
@@ -0,0 +1,46 @@
...
+from mex.extractors.wikidata.helpers import get_wikidata_organization_by_id
+
+
+def get_unit_merged_id_by_synonym(
+    synonym: str,
+) -> MergedOrganizationalUnitIdentifier | None:
+    """Search a merged organizational unit id by synonym.
+
+    Args:
+        extracted_units: list of extracted units
```
MikaMk4: there is no extracted_units argument. is this just a leftover from a refactor?

mex/extractors/organigram/helpers.py
```diff
@@ -0,0 +1,46 @@
...
+        organigram_units,
+        get_extracted_primary_source_id_by_name("organigram"),
+        rki_organization,
+    )
+    unit_merged_ids_by_synonyms = get_extracted_unit_by_synonyms(
+        mex_organizational_units
+    )
+    if synonym not in unit_merged_ids_by_synonyms:
+        return None
+    mex_organizational_unit = unit_merged_ids_by_synonyms[synonym]
```
MikaMk4: there is no logic to meet one of the ACs: if multiple matching units are found it should also return None

MikaMk4 (changes requested):
i am not sure about this first point but the code seems kind of inefficient because the whole process of extracting and transforming units for lookup of a single synonym seems too much. that "setup" for the lookup should be cached itself so that practically only the lookup itself needs to be done when called multiple times.

additionally, the AC requiring the function to be cached seems to also not been met.

mex/extractors/organigram/helpers.py
cutoffthetop: fixed in 7e13b69

cutoffthetop:
> the AC requiring the function to be cached seems to also not been met

true, added that in: dc0d32c

mex/extractors/organigram/helpers.py
cutoffthetop: fair point, but i think the ACs were outdated here. we already raise an error when two units have the same synonym in mex-common: https://github.com/robert-koch-institut/mex-common/blob/1.9.0/mex/common/organigram/extract.py#L76-L81

MikaMk4:
tests are failing

MikaMk4 (approved):

MikaMk4:
👍

PR #105, opened 2025-11-25
MX-1832-ff_projects-creates-organizations

vyvytranngoc:
### PR Context
ff-projects extractor should create organization item if wikidata returns no match

### Added
- get_or_create_organization() for ff_projects in transform.py


esinsj (dismissed):
nice :) approved ✅️

esinsj (approved):

PR #107, opened 2025-11-26
Feature/mx 1790 transform variables from frictionless files

esinsj:
### Added
- transformation of frictionless Metadata.zip files to variables and variable groups


cutoffthetop (approved):
👌

PR #109, opened 2025-12-03
Feature/mx 2025 use organizational units convenience function

erichesse:
### Changes

- use convenience function to obtain organizational unit merged ids


mex/extractors/open_data/transform.py
```diff
@@ -337,11 +334,10 @@ def transform_open_data_parent_resource_to_mex_resource(  # noqa: PLR0913
                     )
                 )
                 for unit_id in unit_list
-                if unit_id not in unit_stable_target_ids_by_synonym[FALLBACK_UNIT]
             }
         )
         if not unit_in_charge:
-            unit_in_charge = unit_stable_target_ids_by_synonym[FALLBACK_UNIT]
```
esinsj: This is specified in the mapping, so it is still needed, or?

mex/extractors/open_data/transform.py
```diff
@@ -337,11 +334,10 @@ def transform_open_data_parent_resource_to_mex_resource(  # noqa: PLR0913
                     )
                 )
                 for unit_id in unit_list
-                if unit_id not in unit_stable_target_ids_by_synonym[FALLBACK_UNIT]
```
esinsj, line 340: This is specified in the mapping, so it is still needed, or?
IT just needs to be changed to 
`if unit_id not in get_unit_merged_id_by_synonym(FALLBACK_UNIT)`

erichesse: This covered the line below, right?

mex/extractors/primary_source/helpers.py
```diff
@@ -1,9 +1,12 @@
+from functools import lru_cache
+
 from mex.common.models import ExtractedPrimarySource
 from mex.common.primary_source.helpers import get_extracted_primary_source_by_name
 from mex.common.types import MergedPrimarySourceIdentifier
 from mex.extractors.sinks import load
 
 
+@lru_cache(maxsize=3)
```
erichesse, line 9: Why 3? There are currently 20 in assets\raw-data\primary-sources\primary-sources.json

erichesse (commented):
Looks good just two comments

mex/extractors/open_data/transform.py
esinsj: the mapping states, that the fallback unit should not be filled in, if there are other units to fill (this is implemented in line 341), but if no units are found at all, then fill in the fallback unit (line 345)

mex/extractors/primary_source/helpers.py
esinsj: "Pi * thumb" 😅
the caching you implemented for organigram has maxize=1. That was too little for primary source, because it jumps between primary sources (e.g. for persons in open data: If you can't find it in the ldap lookup (primary source = ldap), then create a new Person (primary source = open data). 3 was the next best number which looked reasonable to me. But it's absolutely negotiable.

tests/blueant/test_transform.py
```diff
@@ -26,24 +28,18 @@ def test_transform_blueant_sources_to_extracted_activities(
...
         transform_blueant_sources_to_extracted_activities(
             [blueant_source, blueant_source_without_leader],
             stable_target_ids_by_employee_id,
-            unit_stable_target_ids_by_synonym,
             blueant_activity,
             {"Robert Koch-Institut": MergedOrganizationIdentifier.generate(seed=42)},
         )
     )
-    assert len(mex_sources) == 2
+    assert len(mex_sources) == 1
```
esinsj: I don't recognize the changes which would make this test return 1 items instead of 2?

esinsj:
I found two different styles of changing the lookup:
- direct replacement (like in ifsg transform line 75) https://github.com/robert-koch-institut/mex-extractors/pull/109/changes#diff-fd27f84e4bc7655606f6e5731993437b6a0bff7ec18037e85d5a684697549814L75
- creating an own intermediate variable (like in synopse transform line 79) https://github.com/robert-koch-institut/mex-extractors/pull/109/changes#diff-3f6dd9b03de4e4e0be6849033b36df514e440beb48eb0a5b250c92a72cf5828aL79

Is there a reason for this different handling at the different places? Can we unify them?

tests/blueant/test_transform.py
esinsj: fixed in [baf8a5b](https://github.com/robert-koch-institut/mex-extractors/pull/109/commits/baf8a5becda83294d4fb28b90e50190c19fea974)
it's now `len(mex_sources) == 2` again

esinsj (approved):

eckelmannf (approved):
Looks wonderful(l). Trusting in J. E. 🤞

PR #115, opened 2025-12-18
feature/mx-1948 use two stage docker build

MikaMk4:
### Changes
- change the docker build to multi-stage build

tested on a local machine with docker compose:

`docker compose up -d`

needed to change the compose file a bit for the volume mounting to work properly


Dockerfile
```diff
@@ -31,9 +52,7 @@ RUN adduser \
     --uid "10001" \
     mex
 
-COPY . .
-
-RUN --mount=type=cache,target=/root/.cache/pip pip install -r locked-requirements.txt --no-deps
+COPY --chown=mex --exclude=*.lock --exclude=requirements.txt . .
```
cutoffthetop:
```suggestion
COPY --chown=mex assets assets
```

cutoffthetop (approved):

PR #127, opened 2026-01-16
Feature/mx 2148 blueant api update

esinsj:
### PR Context
The API was changed:
- it does not allow anymore to extract all persons. Therefore connector and code were adapted to only retrieve a single person.
- departments can no longer be fetched with the clients/customers endpoint and transformed to MergedOrganizations. This was not wanted anyway, so the clients-request can now return None. The organizations lookup can now handle empty client names. Also the separate lookup parts in extracted.py and transform.py were combined into the one lookup in transform.py 

### Changes
- blueant: fix Person extraction and units written as MergedOrganization items


CHANGELOG.md
```diff
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - lock file maintenance (2026-01-19)
 - remove mailpit path for CI testing
 - datenkompass: publish activities by unit, filter and mapping update
+- blueant: fix Person extraction and units written as MergedOrganization items
```
cutoffthetop: could be in the ### Fixed section then, right? 🙃

cutoffthetop (dismissed):
looks good!

CHANGELOG.md
esinsj: true!
[5e23be5](https://github.com/robert-koch-institut/mex-extractors/pull/127/commits/5e23be5271c77496d54152c930d5a15713432f66)

cutoffthetop (approved):

PR #133, opened 2026-01-19
feature/mx-2120 prep ldap mock

cutoffthetop:
### Added
- prepare testing against a mocked ldap container
  - start mailpit via docker compose for testing
  - remove redundant dagster config from toml (in favor of workspace.yaml)
  - remove redundant dagster defs from `mex.extractors.__init__` (in favor of pointing directly to `load_job_definitions` in the workspace.yaml) -> this prevents triggering a settings load just by importing
  - stop auto-using mocked_igs fixture, in favor of explicit usage -> this prevents unintentional settings loading
  - move OPEN_DATA_EMAIL into constant to make it mockable
  - remove unnecessary `list()` calls from functions that once were generators, but are now just returning lists anyway
  - add mocked unit fixtures, to be used in follow-up branch
  - docker/git ignore .nux folder (dagster runtime artifact)
  - remove deprecated version annotation from compose file


tests/igs/mocked_igs.py
```diff
@@ -11,7 +11,7 @@
 TEST_DATA_DIR = Path(__file__).parent / "test_data"
 
 
-@pytest.fixture(autouse=True)
+@pytest.fixture
```
esinsj, line 14: With this change also the igs test_connector.py integration test needs to be updated, cause it got fixed and gets a different result now.

tests/conftest.py
```diff
@@ -47,3 +55,24 @@ def extracted_organization_rki() -> ExtractedOrganization:
...
+
+
+@pytest.fixture
+def mocked_units_by_identifier_in_primary_source(
+    mocked_extracted_organizational_units: list[ExtractedOrganizationalUnit],
+) -> dict[str, ExtractedOrganizationalUnit]:
+    return {
+        unit.identifierInPrimarySource: unit
+        for unit in mocked_extracted_organizational_units
+    }
```
esinsj, line 78: then the definition of the same two fixtures in open data conftest,py can be deleted, or?

esinsj (changes requested):
looks good to me. just one requested change and one suggestion.

tests/conftest.py
cutoffthetop: fixed in [384a62d](https://github.com/robert-koch-institut/mex-extractors/pull/133/commits/384a62d77fddd4afed19718c5c454bda72cce48f)

tests/igs/mocked_igs.py
cutoffthetop: fixed in [e307241](https://github.com/robert-koch-institut/mex-extractors/pull/133/commits/e307241c22f292d020a7d3dd7e5ae3b73bb0c400)

esinsj (approved):

PR #148, opened 2026-02-03
Feature/mx 1832 ff projects create organization fix

vyvytranngoc:
### Fixed
- ff-projects: fixed extract_ff_projects_organizations() and get_or_create_organization() to match organizations with wikidata


mex/extractors/ff_projects/transform.py
```diff
@@ -121,17 +121,25 @@ def get_or_create_organization(
     """
     final_organizations: list[MergedOrganizationIdentifier] = []
     for org in orgs:
-        if wpo := extracted_organizations.get(org):
+        extracted_organizations = {
+            key.strip().lower(): value for key, value in extracted_organizations.items()
+        }
```
esinsj: this dict comprehension can be moved outside of the loop, so that it isn't computed for each org again. 
But I guess it is necessary here for the idea of expanding _extracted_organizations_? About that see the other comment.

mex/extractors/ff_projects/transform.py
```diff
@@ -121,17 +121,25 @@ def get_or_create_organization(
...
+        new_extracted_organization = ExtractedOrganization(
+            officialName=[Text(value=org)],
+            identifierInPrimarySource=org,
+            hadPrimarySource=get_extracted_primary_source_id_by_name("ff-projects"),
+        )
+        load([new_extracted_organization])
+        merged_id = MergedOrganizationIdentifier(
+            new_extracted_organization.stableTargetId
+        )
+        extracted_organizations[normalized_org] = merged_id
```
esinsj: I understand what you're trying to do, and think it is a good idea. unfortunately, expanding _extracted_organizations_ here doesn't change it for using it next time. Next time it is the original _extracted_organizations_ again.

mex/extractors/ff_projects/transform.py
```diff
@@ -121,17 +121,25 @@ def get_or_create_organization(
...
-                identifierInPrimarySource=org,
-                hadPrimarySource=get_extracted_primary_source_id_by_name("ff-projects"),
-            )
-            load([new_extracted_organization])
-            final_organizations.append(
-                MergedOrganizationIdentifier(new_extracted_organization.stableTargetId)
-            )
+            continue
+        new_extracted_organization = ExtractedOrganization(
+            officialName=[Text(value=org)],
```
esinsj, line 149: I think striping the org name here, before writing the extracted organization is still a good idea.

mex/extractors/ff_projects/transform.py
esinsj: I know I reviewed and approved this last time. But I think using "orgs" within the "for org in orgs" loop doesn't make sense, or?

esinsj (commented):
Nice idea how to solve this. I have some feedback.

mex/extractors/ff_projects/transform.py
vyvytranngoc: I think I misnamed the variable. This is my temporal dict for a single function call. Without updating the dict normalized names (G-BA, G-BA, g-ba) in the same input list would each create a new organization repeatedly. This ensures only one ExtractedOrganization is created and reused during that functioncall

mex/extractors/ff_projects/transform.py
vyvytranngoc: I would keep the original org value. stripping/normalizing org name is only for deduplication and lookup needed. E.g. for original name `"G-BA"` would be stored as `"g-ba"`. But if a want to see whether `"G-Ba"` is already created, I can look up `"g-ba"` which is the stripped and lowercased version :)

mex/extractors/ff_projects/transform.py
esinsj: I understand. I guess it depends on how you resolve the function call in transform.py line 61. If you call it per org, there will only be one org per function call anyway.

mex/extractors/ff_projects/transform.py
esinsj: ah, I meant just stripping, no lower cases. Because there are organizations like " Fancy Company", where one could directly have the clean name without leading or trailing spaces ("Fancy Company") when creating organization.

tests/ff_projects/test_transform.py
```diff
@@ -84,10 +84,23 @@ def test_get_or_create_organization(monkeypatch: MonkeyPatch) -> None:
...
+    ]
     result = get_or_create_organization(org_names, extracted_organizations)
 
-    assert len(result) == 2
+    assert len(created_orgs) == 1
+    assert len(result) == 4
     assert existing_org_id in result
     result_created_orgs = [org_id for org_id in result if org_id != existing_org_id]
     assert result_created_orgs[0] == created_orgs[0].stableTargetId
+    g_ba_ids = [org_id for org_id in result if org_id != existing_org_id]
```
esinsj:
```suggestion
    new_org_ids = [org_id for org_id in result if org_id != existing_org_id]
```

tests/ff_projects/test_transform.py
```diff
@@ -84,10 +84,23 @@ def test_get_or_create_organization(monkeypatch: MonkeyPatch) -> None:
...
+    assert len(set(g_ba_ids)) == 1
```
esinsj:
```suggestion
    assert len(set(new_org_ids)) == 1
```

esinsj (dismissed):
Nice, one tiny last adaptation, and it's good to go.

esinsj (approved):

PR #155, opened 2026-02-13
Feature/mx 2144 datenkompass resources

esinsj:
### Changes

- datenkompass: update resource handling for wave 2

#### Overview of changes:
* add filter-mapping
* models: add additional field
* extract: 
   * add fetching of ExtractedItems
   * improve getting the primary source ids
* filter: 
   * add function to filter by unit 
   * add function to filter for editor primary source
* main: 
   * switch off activities and BibResources for now 
   * load new filter mapping as asset
   * load resources by new filter mapping (less complicated now)
   * add filtered_datenkompass_merged_resources_by_primary_source asset
   * add datenkompass_merged_resources_by_primary_source_by_unit asset
   * adapt functions to new assets-type
   * use new sink with more file name options
* settings: 
   * add primary source name which needs special handling
   * add new variables to handle new keyword concatenation rules
* transform: 
   * add new function to handle new keyword concatenation rules
   * for transfoming resources: move fields with default filling (defined in mapping) out of loop
   * add additional loop to handle new nested dict layer
* sinks: add more file name options
* conftest
   * remove mapping fixture
   * give more meaningful names
   * add another merged resource to be filtered by unit and another merged resource to be filtered by primary source
   * add an extracted resource and let it be returned by mocked_backend
   * make mocked provider handle the added extracted resource
* test_extract: add test for new extraction of extractedItems
* test_filter: add test for new filter functions (by unit, by primary source)
* test_transform:
   * remove uneccessary 'something=None' from testing the lookup for default mappings
   * add test for new function to handle new keyword concatenation rules
   * adapt test to to handle new nested dict layer and more meaningful names


mex/extractors/datenkompass/extract.py
```diff
@@ -38,10 +40,44 @@ def get_merged_items(
...
+    """
+    connector = BackendApiConnector.get()
+
+    response = connector.fetch_extracted_items(
+        entity_type=entity_type,
+        skip=0,
+        limit=1,
+    )
+    total_item_number = response.total
+    item_number_limit = 100  # 100 is the maximum possible number per get-request
```
erichesse, line 67: if this is the default, we don't need to pass the parameter, right?

mex/extractors/datenkompass/extract.py
```diff
@@ -38,10 +40,44 @@ def get_merged_items(
...
+        )
+
+    return [item.stableTargetId for item in extracted_items]
+
+
 def get_filtered_primary_source_ids(
     filtered_primary_sources: list[str] | str | None,
 ) -> list[str]:
-    """Get the IDs of the relevant primary sources.
+    """Get the MergedIdentifier of the relevant primary sources.
```
erichesse:
```suggestion
    """Get a list of MergedIdentifier of filtered primary sources.
```

mex/extractors/datenkompass/extract.py
```diff
@@ -56,16 +92,13 @@ def get_filtered_primary_source_ids(
...
-        str(mps.identifier)
-        for mps in merged_primary_sources
-        if provider.fetch(stable_target_id=mps.identifier)[0].identifierInPrimarySource
-        in filtered_primary_sources
+        mps.stableTargetId
+        for fps in filtered_primary_sources
+        for mps in provider.fetch(
+            identifier_in_primary_source=fps,
+            had_primary_source=MEX_PRIMARY_SOURCE_STABLE_TARGET_ID,
+        )[:1]
```
erichesse: only the first mps? I don't understand that

mex/extractors/datenkompass/filter.py
```diff
@@ -70,6 +73,57 @@ def filter_activities_for_organization_and_unit(
...
+    for filter_unit, allowed_units in allowedunits_by_filterunits.items():
+        result_resources_by_primary_source: dict[str, list[MergedResource]] = {}
+        for (
+            primary_source,
+            merged_resources,
+        ) in merged_resources_by_primary_source.items():
+            allowed_merged_resources = [
+                item
+                for item in merged_resources
+                if allowed_units.intersection(item.unitInCharge)
```
erichesse, line 95: "Only publish MergedResource items, if the items fulfill all of the following requirements.
          The unitInCharge is the unit or a sub-unit of the units mentioned in forValues ."
I'm a little confused about whether this actually reflects the rule

mex/extractors/datenkompass/filter.py
```diff
@@ -102,3 +156,37 @@ def find_descendant_units(
...
+        entity_type: entity type to of merged items
+
+    Settings: primary source which needs to be filtered
+
+    Returns:
+        dictionary with list of filtered merged items
+    """
+    settings = Settings.get()
+
+    primary_source_filter = settings.datenkompass.primary_source_filter
```
erichesse: personally I would rather have this in mappings and not in the settings. Speaking of mapping, I couldn't find such a rule in there

mex/extractors/datenkompass/main.py
```diff
@@ -226,32 +219,39 @@ def datenkompass_merged_resources_by_primary_source_str(
...
-        get_merged_items(
-            entity_type=entity_type,
-            referenced_identifier=datenkompass_filtered_merged_organizational_units_by_id,
-            reference_field="unitInCharge",
-        ),
+@asset(group_name="datenkompass")
+def datenkompass_filtered_merged_resources_by_primary_source(
+    datenkompass_merged_resources_by_primary_source: dict[str, list[MergedResource]],
+) -> dict[str, list[MergedResource]]:
+    """Filter the merged items for primary source as defined in settings."""
```
erichesse: adapt docstring if necessary

erichesse (commented):
A few remarks

mex/extractors/datenkompass/extract.py
esinsj: That is right. I thought because I need it for calculating the range loop, I would like to set it for both. just in case the max default gets changed in backend and we forget to adapt it here.

mex/extractors/datenkompass/main.py
esinsj: [eaccb67](https://github.com/robert-koch-institut/mex-extractors/pull/155/commits/eaccb67544997e3b0b555a9708be3322b5e17ebe)

erichesse (approved):

PR #166, opened 2026-03-13
feature/mx-2166-setup-dev-s3

MikaMk4:
### PR Context
preparation for use of own s3 store in dev environment. we need to tell the s3 client (boto3) to trust our self-signed cert from our s3 store.

### Added

- add setting to control ssl verification for s3 sink

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- add setting to not verify s3 ssl certificate
+- add setting for controlling s3 ssl certificate verification
 
 ### Changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- add setting to not verify s3 ssl certificate
+
 ### Changes
 
 ### Deprecated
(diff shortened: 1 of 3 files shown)


mex/extractors/sinks/s3.py
```diff
@@ -30,6 +30,7 @@ def __init__(self) -> None:
             endpoint_url=str(settings.s3_endpoint_url),
             aws_access_key_id=settings.s3_access_key_id.get_secret_value(),
             aws_secret_access_key=settings.s3_secret_access_key.get_secret_value(),
+            verify=settings.s3_verify_ssl_certificate,
```
rababerladuseladim: I thought our internal S3 will automatically get a certificate, which is verifiable by the rki cert chain. Does it not?

If it does, this setting will need to allow providing a custom certificate, same as [mex-common setting `verify_session`](https://github.com/robert-koch-institut/mex-common/blob/e833dd1a42e93a7ad96f33c55d8f33e6261670c3/mex/common/settings.py#L145)

CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- add setting to not verify s3 ssl certificate
```
rababerladuseladim:
```suggestion
- add setting for controlling s3 ssl certificate verification
```

rababerladuseladim (changes requested):
One question and a suggestion

mex/extractors/sinks/s3.py
MikaMk4: oh yeah you're right. somehow i thought of s3 as it's own protocol but it still is over http/https. then it should work like that!

rababerladuseladim (approved):

PR #169, opened 2026-03-26
feature/mx-2173 open-data: AssetCheck `x_items_less_than`

ZehraVictoria:
### Added

- AssetCheck `x_items_less_than` for open-data

- Pipeline/checks/main.py also had to be adjusted to account for different rule types


mex/extractors/pipeline/checks/main.py
```diff
@@ -133,10 +129,19 @@ def check_item_count_rule(  # noqa: C901
     current_number_of_extracted_items = (
         events[0].asset_materialization.metadata["num_items"].value
     )
-    historical_events = get_historical_events(events)
-    historic_count = get_historic_count(historical_events, time_frame)
-    if historic_count <= 0:
-        return True
+
+    if rule_name in ["less_than_x_inbound", "less_than_x_outbound"]:
+        historic_count = 0
```
erichesse:
```suggestion
    if rule_name in ["less_than_x_inbound", "less_than_x_outbound"]:
        return True
```
and then you don't need to put the lines below into an else, right?

erichesse (approved):

PR #180, opened 2026-04-09
feature/mx-1969-assetcheck-x-percent-less-than

ZehraVictoria:
### Added

- implement x_percent_less_than AssetCheck for all involved extractors

- also removed a few rules from the yamls that weren't implemented yet, too keep it clean


mex/extractors/ff_projects/main.py
```diff
@@ -100,7 +100,12 @@ def ff_projects_activities(
         for ff_projects_source in ff_projects_sources
     ]
     load(extracted_activities)
-    return extracted_activities
+    return Output(
+        value=extracted_activities,
```
esinsj: I'm afraid we can't use it like this. This approach stores the activities list persistently for each daily run for up to 10 month. We already have storage issues, and this would blow up the needed saving space too much.
Do you want to find a possible solution yourself? I was curious and already googled how to solve this 😅 up to you 😁

esinsj (commented):
We need a different solution, due to storage issues

mex/extractors/ff_projects/main.py
ZehraVictoria: The new approach seems to work quite well, the extractors run without problems. I just added one more helper function. Otherwise, changes. [9e1ba4c](https://github.com/robert-koch-institut/mex-extractors/pull/180/commits/9e1ba4c0450eaf6c05f380af53edb8e38063ffc8)

mex/extractors/blueant/main.py
```diff
@@ -82,12 +79,8 @@ def blueant_extracted_activities(
...
     num_items = len(extracted_activities)
     load(extracted_activities)
-    return Output(
-        value=num_items,
-        metadata={
-            "num_items": MetadataValue.int(num_items),
-        },
-    )
+    context.add_output_metadata({"num_items": num_items})
+    return num_items
```
esinsj:
```suggestion
    return extracted_activities
```

I would rather go back to the original (before introduction of asset checks), for consistency reasons. Now that we know that we don't need to return the metadata, let's keep it consistent with all other assets to avoid confusion.

mex/extractors/endnote/main.py
```diff
@@ -74,7 +75,8 @@ def endnote_extracted_bibliographic_resources(
     )
     num_items = len(extracted_bibliographic_resource)
     load(extracted_bibliographic_resource)
-    return Output(value=num_items, metadata={"num_items": MetadataValue.int(num_items)})
+    context.add_output_metadata({"num_items": num_items})
+    return num_items
```
esinsj:
```suggestion
    return extracted_bibliographic_resource
```

same here.

mex/extractors/pipeline/checks/main.py
```diff
@@ -78,6 +81,25 @@ def get_historical_events(events: list[EventLogRecord]) -> dict[datetime, int]:
...
+    latest_materialization = events[0].asset_materialization
+    if latest_materialization is None:
+        return None
+
+    num_items_metadata = latest_materialization.metadata.get("num_items")
+    if num_items_metadata is None or not hasattr(num_items_metadata, "value"):
+        return None
+
+    if num_items_metadata.value is None:
+        return None
```
esinsj: I like that improved the code by moving this into its own function and handle different cases like this.
An additional idea: I think if these cases happen, something went wrong and an error should be raised, or? Otherwise the check will default to "successfully passed"

esinsj (commented):
I like your structured code. Just some add-ons.

mex/extractors/pipeline/checks/main.py
ZehraVictoria: [806306f](https://github.com/robert-koch-institut/mex-extractors/pull/180/commits/806306f0d4af529294b0354561baf2e727ab631e) like this?

mex/extractors/pipeline/checks/main.py
```diff
@@ -81,21 +84,24 @@ def get_historical_events(events: Sequence[EventLogRecord]) -> dict[datetime, in
...
+    num_items_metadata = (
+        latest_materialization.metadata.get("num_items")
+        if latest_materialization is not None
+        else None
+    )
+    if (
+        latest_materialization is None
+        or num_items_metadata is None
+        or not hasattr(num_items_metadata, "value")
+        or num_items_metadata.value is None
```
esinsj: now it is a bit too much 😅
E.g. or the first asset check or the first job run there won't be any metadata.
I meant an error should be raised if
```
not hasattr(num_items_metadata, "value") or
num_items_metadata.value is None
```
that shouldn't happen. But no latest materialization or no metadata can happen.

esinsj (commented):
small misunderstanding

mex/extractors/pipeline/checks/main.py
ZehraVictoria: Ok, yeah that makes sense. Sorry, I was a bit confused [741a1b3](https://github.com/robert-koch-institut/mex-extractors/pull/180/commits/741a1b34bda44428edb3494f5cceb511db897c54)

esinsj (approved):
nice :) approved

PR #199, opened 2026-05-07
feature/mx-1967 less_than_x_outbound AssetCheck

ZehraVictoria:
### Added

- implement "less_than_x_outbound" AssetCheck

- since this check is currently only implemented for synopse, I wasn't sure where to put the helper count_outbound_connections (utils.py or new helpers.py for synapse?). It also only works for ExtractedVariable for now


mex/extractors/pipeline/checks/main.py
```diff
@@ -93,7 +95,11 @@ def get_latest_num_items(events: Sequence[EventLogRecord]) -> int | None:
     if latest_materialization is None:
         return None
 
-    num_items_metadata = latest_materialization.metadata.get("num_items")
+    if rule_name == "less_than_x_outbound":
```
erichesse, line 98: I was thinking, we could use the `STATIC_RULES`, but that doesn't work right?

erichesse (approved):
I feel like the rule implementation could be more elegant. But the suggestions I came up with, aren't convincing me neither. Up to you whether you want to merge or adapt something

mex/extractors/pipeline/checks/main.py
ZehraVictoria: All three `STATIC_RULES` need different items from the metadata. E.g.` not_exactly_x_items ` need `num_items` so I just don't see a way of getting around having 3 separate if cases for each rule here. I had a look at restructuring the code, but I feel like this only move the problems to different places.
Maybe I can think of something good once I implemented the last static rule `less_than_x_inbound` and I have a complete view of all the functions.
For now, I'm also not completely happy with the code, but I'm ok with merging.

PR #200, opened 2026-05-08
feature/mx_1792_seq_repo_update

erichesse:
### Changes

- seq-repo mapping update


esinsj (approved):
pipeline doesn't run, but that's a mapping issue. otherwise it looks good

PR #235, opened 2026-06-15
feature/mx-2264_update_to_model_v5.0_numberOf

esinsj:
### Added
- extraction of _numberOfRecords_ and _numberOfUniqueIndividuals_

### Removed
- extraction of _sizeOfDatabase_


mex/extractors/grippeweb/transform.py
```diff
@@ -181,7 +180,6 @@ def transform_grippeweb_resource_mappings_to_dict(
             resourceTypeGeneral=resource_type_general,
             resourceTypeSpecific=resource_type_specific,
             rights=rights,
-            sizeOfDataBasis=size_of_data_basis,
```
erichesse, line 184: was this overlooked in the mapping PR?

mex/extractors/biospecimen/models/source.py
```diff
@@ -27,7 +27,7 @@ class BiospecimenResource(BaseModel):
     ressourcentyp_allgemein: str | None
     ressourcentyp_speziell: list[str]
     rechte: str | None
-    vorhandene_anzahl_der_proben: str | None
```
erichesse, line 30: There can be strings in these cells. `uv run biospecimen` terminates with an error

erichesse (commented):
Looks good, just two remarks

mex/extractors/grippeweb/transform.py
esinsj: I verified with mdmgmt that _sizeOfDataBasis_ is deleted without any replacement for grippeweb. I forgot to forward that information.

mex/extractors/biospecimen/models/source.py
esinsj: you need to check out branch mx-2309 in assets. But I don't think you can run the extractors in the single field-branches just yet. We'll have to wait until each field-branch is merged into the feature branch, because the assets-branch contains all changes, and the single field-branches each contain only a part of it.

erichesse (approved):

PR #242, opened 2026-06-17
Feature/mx 2249 update to model v5.0 relatedResource

esinsj:
### Added
- extraction of _relatedResource_

### Removed
- extraction of _isPartOf_


erichesse (approved):
Looks good!

PR #246, opened 2026-06-18
Feature/1966 assetcheck inbound connections

ZehraVictoria:
### Added

- implement less_than_x_inbound AssetCheck
- pipeline now allows two rules of the same type per entity type (with different target types), this allows migration of inbound resource check to outbound variable check


ZehraVictoria:
I have not yet cleaned things up for the output connections (I thought about unifiying/streamlining the functions a bit for input/output checks e.g. metadata type), since I still need to go back to the output check. The opendata extractor will also be using the output check, but the check has not been implemented yet to work for distributions (See Jira Ticket comment).
So I was wondering how to approach this? Should I create a new followup ticket for cleanup and opendata, do a hotfix or use this PR?

mex/extractors/utils.py
```diff
@@ -26,3 +27,27 @@ def count_outbound_connections(variable: ExtractedVariable) -> int:
...
+
+            if not isinstance(related_values, list):
+                related_values = [related_values]
+
+            for related_value in related_values:
+                if related_value is None:
+                    continue
+                identifiers.append(str(related_value))
+
+    return identifiers
```
esinsj, line 38: why not return a dict of the amount of references per identifier (like for outbound)?

mex/extractors/pipeline/checks/main.py
```diff
@@ -151,47 +159,50 @@ def get_historic_count(
...
+        # fail if current number of extracted items is not equal to threshold
+        if current_number_of_items == threshold:
             return True
         return True  # TODO @MX-2298: revert to returning the result of the comparison
     if rule_name == "less_than_x_inbound":
-        pass
+        # fail if any of the inbound connection counts is smaller than threshold
+        if not isinstance(current_number_of_items, list):
+            raise ValueError(LATEST_NUM_ITEMS_ERROR)
+        return len(current_number_of_items) >= threshold
```
esinsj: Unless I overlooked something, I think this counts the amount of all referenced identifiers, not the amount of references per identifier?

esinsj:
My thoughts on

> I have not yet cleaned things up for the output connections (I thought about unifiying/streamlining the functions a bit for input/output checks e.g. metadata type), since I still need to go back to the output check. The opendata extractor will also be using the output check, but the check has not been implemented yet to work for distributions (See Jira Ticket comment). So I was wondering how to approach this? Should I create a new followup ticket for cleanup and opendata, do a hotfix or use this PR?

Since this seems to be a bigger issue, I'd opt for a new ticket. But that depends on the final decision on how to handle this. But a new ticket would make the traceability easier :)

mex/extractors/utils.py
esinsj: I think only _variable.belongsTo_ references VariableGroups (as defined in the check), the other references point to resources and primarysources, respectively

ZehraVictoria: https://github.com/robert-koch-institut/mex-extractors/pull/246/commits/527b3c8cb95e1d324be9f40a39f1ea2d2b26d0e2 Yes, these absolutely need to go!

mex/extractors/pipeline/checks/main.py
ZehraVictoria: Since there is a lot of variability between the different amounts of references to resources (reaching from 31 to >2000 per resource), Franziska decided to accumulate all inbound references. Currently the value in the check is 100, but I will create a new PR in mex-assets to change the value to the current actual sum (>9000). I would also like to do that in the already mentioned new ticket, if thats ok for you?

mex/extractors/utils.py
ZehraVictoria: Good idea [7c11a16](https://github.com/robert-koch-institut/mex-extractors/pull/246/commits/7c11a16d9fcd37ecdee4f19ef22ebf249e41bb70)

assets/raw-data/pipeline/voxco/resource.yaml
```diff
@@ -3,3 +3,6 @@
 rules:
   - fail_if: not_exactly_x_items
     value: 2
+  - fail_if: less_than_x_inbound
+    value: 100
+    target_type: Variable
```
esinsj: Quick question if I see it correctly:
synopse has an outbound rule: variable pointing at resource.
voxco has an inbound rule: resource being pointed at by variable.

would it be possible to convert all inbound rules to outbound rules and ditch the inbound rules altogether? I have the feeling that would make the whole thing easier?

mex/extractors/pipeline/checks/main.py
```diff
@@ -85,20 +96,32 @@ def get_historical_events(events: Sequence[EventLogRecord]) -> dict[datetime, in
...
     if latest_materialization is None:
         return None
 
-    if rule_name == "less_than_x_outbound":
-        num_items_metadata = latest_materialization.metadata.get("outbound_connections")
-    else:
-        num_items_metadata = latest_materialization.metadata.get("num_items")
+    metadata_key_by_rule = {
+        ("less_than_x_inbound", None): "inbound_connections",
+        ("less_than_x_outbound", None): "outbound_connections",
```
esinsj: we could give the target type as a key for all outbound rules, for consistency and then we could also do without the ("less_than_x_outbound", None). And if we get rid of the inbound rule, also without the ("less_than_x_inbound", None)

esinsj (commented):
not completely done with the review yet, but I have an idea I wanted to share.

assets/raw-data/pipeline/voxco/resource.yaml
ZehraVictoria: Already talked with Franziska about that and planned to maybe do that in a follow up ticket. 
But I actually don't see why we should wait - she's also fine with kicking the inbound checks out completely.
[0b214e6](https://github.com/robert-koch-institut/mex-extractors/pull/246/commits/0b214e612f4db48648eb0f69a28732cc392fdacd)

mex/extractors/pipeline/checks/main.py
ZehraVictoria: [480e33d](https://github.com/robert-koch-institut/mex-extractors/pull/246/commits/480e33d62f3dea357c78d0849351d602d64c55ef)

mex/extractors/pipeline/checks/main.py
```diff
@@ -151,47 +174,48 @@ def get_historic_count(
...
 
     if rule_name == "not_exactly_x_items":
-        if current_number_of_extracted_items == threshold:
+        # fail if current number of extracted items is not equal to threshold
+        if current_number_of_items == threshold:
             return True
         return True  # TODO @MX-2298: revert to returning the result of the comparison
-    if rule_name == "less_than_x_inbound":
-        pass
+        return sum(current_number_of_items.values()) >= threshold
```
esinsj: I think this line slipped through the delition in 0b214e612f4db48648eb0f69a28732cc392fdacd ?

mex/extractors/pipeline/checks/main.py
```diff
@@ -251,16 +277,16 @@ def check_item_count_rule(
...
         return True
 
     # Handle static rules
     if rule_name in STATIC_RULES:
-        if not check_static_rule(rule_name, current_number_of_extracted_items, rule):
+        if not check_static_rule(rule_name, current_number_of_items, rule):
             msg = (
                 f"Asset {asset_key} failed {rule_name} check: "
-                f"{current_number_of_extracted_items} not meeting threshold."
+                f"{current_number_of_items} not meeting threshold."
```
esinsj:
```suggestion
                f"{current_number_of_items} not meeting threshold of {rule['value']}."
```

esinsj: just an idea for a more telling error message. Not sure if it might be more confusing than helpful.

esinsj (commented):
Only some minor changes. otherwise it looks good

mex/extractors/pipeline/checks/main.py
ZehraVictoria: no, that makes sense!

mex/extractors/pipeline/checks/main.py
ZehraVictoria: Upsi, deleted in [93a2e0d](https://github.com/robert-koch-institut/mex-extractors/pull/246/commits/93a2e0d0330893f7dbe67dab26660a59b5ee7a27)

esinsj (approved):

PR #255, opened 2026-06-24
Feature/bugfix add mocked wikidata for ldap convience function usage

esinsj:
### Fixed
- add wikidata-fixture to some extractor test


tests/blueant/test_extract.py
```diff
@@ -17,7 +17,7 @@ def test_extract_blueant_sources_from_api() -> None:
     assert sources[0].model_fields_set == BlueAntSource.model_fields.keys()
 
 
-@pytest.mark.usefixtures("mocked_ldap", "mocked_blueant")
```
esinsj, line 20: the ldap fixture isn't needed here, so I cleaned it up.

rababerladuseladim (approved):

PR #258, opened 2026-06-25
feature/mx_2308_seq_repo_fallback

erichesse:
### Changes
- implement seq-repo fallback rule for units

### DIFF
--- a/mex/extractors/seq_repo/transform.py
+++ b/mex/extractors/seq_repo/transform.py
@@ -7,7 +7,6 @@ from mex.common.models import (
     ExtractedAccessPlatform,
     ExtractedActivity,
     ExtractedOrganization,
-    ExtractedPerson,
     ExtractedResource,
     ResourceMapping,
 )
@@ -22,6 +21,8 @@ from mex.extractors.organigram.helpers import get_unit_merged_id_by_synonym
 from mex.extractors.primary_source.helpers import (
     get_extracted_primary_source_id_by_name,
 )
+from mex.extractors.seq_repo.extract import extract_source_project_coordinator_by_name
+from mex.extractors.settings import Settings
 
 if TYPE_CHECKING:
     from mex.extractors.seq_repo.model import SeqRepoSource
@@ -30,15 +31,12 @@ if TYPE_CHECKING:
 def transform_seq_repo_activities_to_extracted_activities(
     seq_repo_sources: list[SeqRepoSource],
     seq_repo_activity: ActivityMapping,
-    seq_repo_extracted_persons_by_name: dict[str, ExtractedPerson],
 ) -> list[ExtractedActivity]:
     """Transform seq-repo activities to list of unique ExtractedActivity.
 
     Args:
         seq_repo_sources: Seq Repo extracted sources
         seq_repo_activity: Seq Repo activity mapping models with default values
-        seq_repo_extracted_persons_by_name: Seq Repo sources resolved project
-                                            coordinators ldap query results
 
     Returns:
         list of unique ExtractedActivity
@@ -48,8 +46,7 @@ def transform_seq_repo_activities_to_extracted_activities(
 
     for source in seq_repo_sources:
         contact, responsible_units = get_resolved_project_coordinators_and_units(
-            source.project_coordinators,
-            seq_repo_extracted_persons_by_name,
+            source.project_coordinators, seq_repo_sources
         )
         involved_person = [c for c in contact if isinstance(c, MergedPersonIdentifier)]
         extracted_activity = ExtractedActivity(
@@ -68,12 +65,11 @@ def transform_seq_repo_activities_to_extracted_activities(
     return unique_activities
 
 
-def transform_seq_repo_resource_to_extracted_resource(  # noqa: PLR0913
+def transform_seq_repo_resource_to_extracted_resource(
     seq_repo_sources: list[SeqRepoSource],
     seq_repo_activities: dict[str, ExtractedActivity],
     mex_access_platform: ExtractedAccessPlatform,
     seq_repo_resource: ResourceMapping,
-    seq_repo_extracted_persons_by_name: dict[str, ExtractedPerson],
     extracted_organization_rki: ExtractedOrganization,
 ) -> list[ExtractedResource]:
     """Transform seq-repo resources to ExtractedResource.
@@ -83,8 +79,6 @@ def transform_seq_repo_resource_to_extracted_resource(  # noqa: PLR0913
         seq_repo_activities: Seq Repo extracted activity for default values from mapping
         mex_access_platform: Extracted access platform
         seq_repo_resource: Seq Repo resource mapping model with default values
-        seq_repo_extracted_persons_by_name: Seq Repo Sources resolved project
-                                            coordinators as Extracted Persons
         extracted_organization_rki: wikidata extracted organization
 
     Returns:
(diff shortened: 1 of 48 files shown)


mex/extractors/seq_repo/transform.py
```diff
@@ -266,7 +259,9 @@ def get_resolved_project_coordinators_and_units(
     Returns:
         Resolved ids project coordinator and units
     """
-    project_coordinators_ids: set[MergedPersonIdentifier] = set()
+    project_coordinators_ids: set[
+        MergedPersonIdentifier | MergedOrganizationalUnitIdentifier
+    ] = set()
     units_in_charge: set[MergedOrganizationalUnitIdentifier] = set()
     for pc in project_coordinators:
         person = seq_repo_extracted_persons_by_name.get(pc)
```
esinsj: _seq_repo_extracted_persons_by_name_ only contains extracted Persons, but some of the listed "project-coordinators" can also be functional accounts, which need to be matched for Activity contact and Resource contact.

esinsj (commented):
one remark

mex/extractors/seq_repo/transform.py
erichesse: https://github.com/robert-koch-institut/mex-extractors/pull/258/commits/02516f784eca110a805f6df34e49dcb64f8665bd

mex/extractors/seq_repo/transform.py
```diff
@@ -266,14 +265,34 @@ def get_resolved_project_coordinators_and_units(
...
+                person.stableTargetId or contact_id,
+            )
+        )
         if unit := person.memberOf:
             units_in_charge.update(unit)
 
+    if not units_in_charge:
+        units_in_charge = cast(
+            "set[MergedOrganizationalUnitIdentifier]",
+            get_unit_merged_id_by_synonym("mf1"),
```
esinsj: I think it's better to not hardcode the fallback here but to put it into a setting variable. Otherwise we'll have to add it to our test data, and that mixes up things. If it's in a variable, we can give another unit for testing.

mex/extractors/seq_repo/transform.py
```diff
@@ -266,14 +265,34 @@ def get_resolved_project_coordinators_and_units(
...
+    project_coordinators_ids: set[
+        MergedContactPointIdentifier
+        | MergedPersonIdentifier
+        | MergedOrganizationalUnitIdentifier
+    ] = set()
     units_in_charge: set[MergedOrganizationalUnitIdentifier] = set()
     for pc in project_coordinators:
         person = seq_repo_extracted_persons_by_name.get(pc)
-        if not person:
+        contact_id = get_ldap_merged_contact_id_by_mail(mail=pc)
```
esinsj: Idea: This part could also be looked up in extract.py where _seq_repo_extracted_persons_by_name_ is created. Then we could also avoid loading the same contact point several times, by also making use of the seen functionality. Or get rid of that _seq_repo_extracted_persons_by_name_ asset altogether and do both lookups (persons and contact points) right here.

esinsj: I'm also not sure, if the functional accoutns are given as email-adress. At least not according to the mapping in the example.

esinsj (commented):
some more ideas

mex/extractors/seq_repo/transform.py
erichesse: https://github.com/robert-koch-institut/mex-extractors/pull/258/commits/f27ed368ee95a5faa934dac9be85d17bb980fa12
done

mex/extractors/seq_repo/transform.py
erichesse: I pass it as mail now(in consent with mdmgmt), I tested it and it worked
https://github.com/robert-koch-institut/mex-extractors/pull/258/commits/f27ed368ee95a5faa934dac9be85d17bb980fa12

erichesse: I would like to keep extract and transform logic separated as much as possible. Can be quite difficult to untangle however...
I cut out some überflüssiger code out though, what do you think?
https://github.com/robert-koch-institut/mex-extractors/pull/258/commits/f27ed368ee95a5faa934dac9be85d17bb980fa12

mex/extractors/seq_repo/transform.py
```diff
@@ -254,26 +239,56 @@ def transform_seq_repo_access_platform_to_extracted_access_platform(
...
         project_coordinators: Seq Repo raw project coordinator names
-        seq_repo_extracted_persons_by_name: Seq Repo Sources resolved project
-                                            coordinators as Extracted Persons
+        seq_repo_sources: Seq Repo extracted sources
 
     Returns:
         Resolved ids project coordinator and units
     """
-    project_coordinators_ids: set[MergedPersonIdentifier] = set()
+    seq_repo_extracted_persons_by_name = extract_source_project_coordinator_by_name(
```
esinsj: Calling the function here would execute it once per resource, so roughly 60,000 times. If we keep this logic, it would be better to access the precomputed value via the asset again instead of recalculating it in the loop.

esinsj (commented):
sorry, I found a disadvantage of the partial solution.

mex/extractors/seq_repo/transform.py
erichesse: https://github.com/robert-koch-institut/mex-extractors/pull/258/commits/30c42373eab15d698bc0c35a0e4be7f0481399fa

erichesse: I log the resources directly now

esinsj: Thanks for revisiting this. I still have some concerns about calling a function, that fetches the persons for all resources, once for each resource. Although this is no longer visible in the logs, the reduced logging seems to depend partly on the caching behavior of the LDAP helper.

With the current implementation, I’m also not sure whether the "seen"-logic in `extract_source_project_coordinator_by_name` provides additional value. Unless I’m missing something, this could mean that the helper function itself is no longer necessary. One possible simplification would be to then directly call
`person = get_ldap_extracted_person_by_query(sam_account_name=pc)` in transform.py at line 274.
Alternatively, reintroducing the `seq_repo_extracted_persons_by_name` asset in main could be an option.

What are your thoughts on these two options?

mex/extractors/seq_repo/transform.py
```diff
@@ -254,26 +242,56 @@ def transform_seq_repo_access_platform_to_extracted_access_platform(
...
+        MergedContactPointIdentifier
+        | MergedPersonIdentifier
+        | MergedOrganizationalUnitIdentifier
+    ] = set()
     units_in_charge: set[MergedOrganizationalUnitIdentifier] = set()
     for pc in project_coordinators:
         person = seq_repo_extracted_persons_by_name.get(pc)
-        if not person:
+        contact_id = get_ldap_merged_contact_id_by_mail(mail=f"{pc}@rki.de")
+        if not person or contact_id:
```
esinsj:
```suggestion
        if not (person or contact_id):
```

mex/extractors/seq_repo/transform.py
erichesse: https://github.com/robert-koch-institut/mex-extractors/pull/258/commits/1952cec40a15c20d66fa73bf476831ccc4b06ce0

esinsj (approved):

PR #295, opened 2026-07-16
Feature/mx 2277 add csv export to publishing pipeline

esinsj:
### Added
- publish CSV of publications to be downloadable on MEx


mex/extractors/publisher/extract.py
```diff
@@ -21,3 +22,28 @@ def get_publishable_merged_items(
         reference_filters=reference_filters,
     )
     return list(response)
+
+
+def get_publishable_merged_item_by_identifier(
+    identifier: str,
+) -> AnyMergedModel:
```
erichesse:
```suggestion
def get_publishable_merged_item(
    identifier: str,
) -> AnyMergedModel:
```
we use "by" usually for dicts. Could you also expand the docstring including raised errors?

mex/extractors/publisher/filter.py
```diff
@@ -52,3 +67,114 @@ def filter_persons_with_approving_unique_consent(
...
+            for unit_name in activity_filter_rule_by_field["responsibleUnit"]
+            .filterRules[1]
+            .forValues
+            for unit_id in (get_unit_merged_id_by_synonym(unit_name) or [])
+        }
+    else:
+        msg = "No units set for __all__/activity_filter.yaml"
+        raise MExError(msg)
+
+    return forbidden_unit_ids
```
erichesse, line 94: could `filter_by_global_rules` be used instead? That is the general way to filter by `__all__` rules.

mex/extractors/publisher/filter.py
```diff
@@ -52,3 +67,114 @@ def filter_persons_with_approving_unique_consent(
...
+        bibliographic_resource_by_department[department_unit_id] = (
+            _cluster_publications_by_department(
+                department_unit_id,
+                merged_organisational_units,
+                forbidden_unit_ids,
+                merged_bibliographic_resources,
+            )
+        )
+
+    return bibliographic_resource_by_department
```
erichesse, line 180: My idea would be to do something like:
`filtered_merged_bibliographic_resources = filter_by_global_rules (merged_bibliographic_resources)`
Where you would probably need a model that in inherits from `MergedBibliographicResource` and implements the respective methods.
and then do a for loop where each unit is added to to the value list for the respective department key. Does this make sense?

mex/extractors/publisher/main.py
```diff
@@ -184,6 +192,48 @@ def publisher_sink_load(publisher_items: PublisherItemsLike) -> None:
...
 
+@asset(group_name="publisher")
+def publisher_bibliographic_resources_for_csv_by_unit() -> dict[
+    str, list[BibliographicResourceForCsv]
+]:
+    """Extract the bibliographic resources and transform to format for the csv list."""
+    merged_bibliographic_resources = cast(
+        "list[MergedBibliographicResource]",
+        get_publishable_merged_items(entity_type=["MergedBibliographicResource"]),
+    )
```
erichesse, line 203: I would factor that part out of main

mex/extractors/publisher/settings.py
```diff
@@ -14,6 +15,17 @@ class PublisherSettings(BaseModel):
...
+        AssetsPath("mappings"),
+        description=(
+            "Path to the directory with the filter mapping files containing the "
+            "default values, absolute path or relative to `assets_dir`."
+        ),
+    )
+    departments_for_publications_csv: list[str] = Field(
+        ["PRNT", "fg99"],
+        description="list of departments for which to publish publications.",
+    )
```
erichesse, line 28: I don't understand why this setting exists. I think it makes more sense to handle this in assets

mex/extractors/publisher/transform.py
```diff
@@ -109,3 +115,104 @@ def update_actor_references_where_needed(
...
+        name as string or None if not found.
+
+    Raises:
+        MExError if name or merged item have the wrong type.
+    """
+    merged_item = get_publishable_merged_item_by_identifier(identifier)
+
+    if field not in merged_item.model_fields:
+        msg = f"Unknown field: '{field}' for {merged_item.entityType}"
+        raise ValueError(msg)
```
erichesse: Do we need to catch this here? Won't this be handled by the backend?

mex/extractors/sinks/s3.py
```diff
@@ -191,3 +162,72 @@ def load(
...
+            unit_name: unit name for csv naming
+
+        Returns:
+            Generator for the loaded items
+        """
+        settings = ExtractorsSettings.get()
+
+        if unit_name is None:
+            msg = "No Unit Name provided for loading publications."
+            raise RuntimeError(msg)
```
erichesse: It's irritating that the default value throws an error. make unit_name a positional argument?

mex/extractors/sinks/s3.py
```diff
@@ -191,3 +162,72 @@ def load(
...
+
+        unitname = unit_name.replace(" ", "")
+        directory_path = build_directory_path("downloadable files")
+
+        publications_file_name = f"Publications_{unitname}.csv"
+        publications_path = (directory_path / publications_file_name).as_posix()
+
+        df = pd.DataFrame(
```
erichesse: I am not sure, why we need a DataFrame here. They do silent type coercion which can cause a lot of trouble. I think loading the dict to csv should work as well

mex/extractors/sinks/write_metadata.py
```diff
@@ -0,0 +1,54 @@
...
+from mex.common.types import UTC
+
+if TYPE_CHECKING:
+    from io import BytesIO, StringIO
+
+
+def build_directory_path(prefix: str | None) -> Path:
+    """Build directory path that includes the mex-model major and minor version."""
+    if prefix is None:
+        prefix = ""
```
erichesse:
```suggestion
def build_directory_path(prefix: str = "") -> Path:
    """Build directory path that includes the mex-model major and minor version."""

```

mex/extractors/sinks/write_metadata.py
```diff
@@ -0,0 +1,54 @@
...
+    mex_model_version = Version(metadata.version("mex-model"))
+    return Path(f"{prefix}-{mex_model_version.major}.{mex_model_version.minor}")
```
erichesse, line 21: if prefix is "", the file starts with -

CHANGELOG.md
```diff
@@ -65,6 +65,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [1.18.0] - 2026-08-03
 
+### Added
+
+- link igs to seq-repo
+
```
erichesse, line 83: merge issue

CHANGELOG.md
```diff
@@ -94,6 +98,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changes
 
 - BREAKING: Setting is renamed to ExtractorsSettings, new env prefix is MEX_EXTRACTORS_
+- updated synopse models according to mapping + CSV and renamed fields in snake_case
```
erichesse, line 113: merge issue

mex/extractors/sinks/s3.py
```diff
@@ -191,3 +162,72 @@ def load(
...
+
+class S3CsvSink(S3BaseSink):
+    """Special sink to load models as CSV file into S3 bucket and publish metadata."""
+
+    def load(
+        self,
+        items_sorted_by_year: Iterable[_LoadItemT],
+        *,
+        unit_name: str | None = None,
+    ) -> Generator[_LoadItemT]:
```
erichesse, line 172: why do we need a return value here?

mex/extractors/publisher/filter.py
esinsj: As far as I understand, that filter is originally for Activities. E.g. it contains fields, that are not present in BibliographicResources. Especially the unit references are called "responsibleUnit" in the filter and "contributingUnit" in BibliographicResources. also, we want to filter only for units, not for the other properties addressed in the activity_filter.yaml.

mex/extractors/publisher/filter.py
esinsj: see other comment

mex/extractors/publisher/main.py
esinsj: ah, I thought the extraction is generally handled inmain. where would you rather put it?

mex/extractors/publisher/extract.py
esinsj: ah, I oriented my self on the name "get_unit_merged_id_by_synonym". I changed it: [0a04c10](https://github.com/robert-koch-institut/mex-extractors/pull/295/commits/0a04c1041c0e2f48177f4e0bebf3fd4cb1b1194a)

mex/extractors/publisher/settings.py
esinsj: I don't understand what you mean by "handle in assets"? do you mean to create a mapping for publisher?

mex/extractors/publisher/transform.py
esinsj: I don't understand your question. 

According to the ticket I need to resolve the names (shortName, officialName) of references items (like unit or organization). 
This function accepts the ID of the referenced item and the field of which I want the content (e.g. "shortName"). The function first uses the ID to get the item (of any entity type) from backend. Once it has the item it tries to get the content of the field. The backend is not involved at that step anymore.

Does it answer your question or could you specify?

mex/extractors/sinks/s3.py
esinsj: I used the BaseSink as a guide, where the load function returns a Generator: https://github.com/robert-koch-institut/mex-common/blob/main/mex/common/sinks/base.py

CHANGELOG.md
esinsj: that is actually the correction of an earlier merge issue.

CHANGELOG.md
esinsj: also an correction of an earlier merge issue

mex/extractors/sinks/write_metadata.py
esinsj: good point! I made the argument an requirement: [baedec2](https://github.com/robert-koch-institut/mex-extractors/pull/295/commits/baedec27b2ebc8a5047f1db9b19cd39bb1aaaa03)

mex/extractors/sinks/write_metadata.py
esinsj: I decided to make the prefix mandatory (see other comment)

mex/extractors/sinks/s3.py
esinsj: good question. I found out that the whole inheritance is not according to best practice and improved it in [042761a](https://github.com/robert-koch-institut/mex-extractors/pull/295/commits/042761adc7615aa66248771403d34061f0b3e018) + [042761a](https://github.com/robert-koch-institut/mex-extractors/pull/295/commits/042761adc7615aa66248771403d34061f0b3e018) + [019851b](https://github.com/robert-koch-institut/mex-extractors/pull/295/commits/019851b62eb1e6b49d727776037ad2f90038fee3)

mex/extractors/sinks/s3.py
esinsj: ok: [5caa289](https://github.com/robert-koch-institut/mex-extractors/pull/295/commits/5caa289efd7bee399b19cc0174c6c01b3501abc3)

mex/extractors/publisher/transform.py
esinsj: [7c6d361](https://github.com/robert-koch-institut/mex-extractors/pull/295/commits/7c6d361a4fc340aa8f7393040d6ee83b7a213ac1)

erichesse (approved):

PR #324, opened 2026-08-03
feature/mx-1142-test-backend-integration

rababerladuseladim:
### Added

- tests: integration tests with mex-backend in CI. Skipped locally to avoid manual
  backend setup, will be automated in MX-1523.

### Changes

- tests: make `mex.bat test` run tests marked with `requires_rki_infrastructure`

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -84,6 +84,10 @@ jobs:
           MEX_BACKEND_API_KEY: write_key
         run: make test
 
+      - name: Dump container logs on failure
+        if: failure()
+        run: docker compose logs mex-backend neo4j
+
       - name: Cleanup containers
         if: always()
         run: docker compose down slapd mailpit neo4j mex-backend
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -36,7 +36,7 @@ jobs:
           fetch-depth: 1
 
       - name: Start containers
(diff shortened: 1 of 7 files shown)


.github/workflows/testing.yml
```diff
@@ -68,12 +68,26 @@ jobs:
...
+            curl -sf http://localhost:8080/v0/_system/check && exit 0
+            sleep 2
+          done
+          echo "mex-backend did not become ready in time" >&2
+          exit 1
+
       - name: Run test suite
         env:
           MEX_LDAP_URL: ldap://cn=admin,dc=ldapmock,dc=local:adminpass@localhost:636
-#          MEX_LDAP_SEARCH_BASE: dc=ldapmock,dc=local
```
erichesse, line 74: Why delete? 
I thought we want enable the ldap mock server eventually like we did in common: https://github.com/robert-koch-institut/mex-common/blob/main/.github/workflows/testing.yml

erichesse (commented):
bty with a question

.github/workflows/testing.yml
rababerladuseladim: Yes, we will. But this was commented out, without any explanation why it was commented out. I was confused why this was here at all and hence removed it to avoid future confusion. When we activate slapd, we can easily introduce it again.

erichesse (approved):

